Issue #182, opened 2021-01-12
Bug in drop_miss implementation


kylebaron:
# Summary

As a user, I want `pt_data_inventory` to drop the Missing Observation summary when I pass `drop_miss = TRUE`

# Tests
- tests/testthat/test-inventory-table.R

PR #183, opened 2021-01-12
fix drop_miss bug

kylebaron:
Issue #182 

- should be simple fix
- simplified logic; will just drop on demand, no checking
-  test added


kylebaron:
Taking Katherine's approval on this one.

Issue #171, opened 2020-12-14
Check for missing values in categorical covariates


kylebaron:
# Summary

As a user, I want pmtables to tell me when there is missing values in categorical covariate columns when creating categorical pmtable and replace with character NA.

# Tests
- tests/testthat/test-discrete-table.R
  - cat table with missing value

PR #184, opened 2021-01-13
Replace missing values with 'NA'

kylebaron:
Issue #171 

- Add helper to prepare categorical data frame (some of this code was getting called repeatedly; now just once)
- If there are missing values in a categorical data column, replace them with <character> NA and warn
- If the column was a factor, then re-create the factor with possibly a new level of "NA"


kylebaron:
Will merge this at the end of day; open for further comment in the meantime.

PR #189, opened 2021-01-23
adds .list argument to st_rename

kylebaron:
# Summary
- add. `.list` argument to `st_rename()`, allowing the user to pass in a list of rename data
- `.list` trumps `...` [in response to reviewer feedback]
- duplicate renames are removed
- the `.list` list should be in `old-name = new-name` format; this is to be compatible with yspec functionality that returns a list named by the data set column and with data as the `short` name (e.g. `list(WT = "weight")`
  - This was initially re-factored based on review feedback to `new = old`; we reverted to the original `old = new` after further 
    discussion with the conclusion that the convenience would be preferred here
  - Added a check of names with warning if no matching names are found 
  - Added deeper explanation of behavior in the docs and added illustrative example 
  - Will consider `.dots` argument in the future too
- I'm going to hold off on using `list2` since that seems (?) to break current behavior; would be happy to re-visit this if there is a more efficient way to get this done with `rlang` stuff
- Added tests

# Examples
```r
st_new(stdata()) %>% st_rename(weight = WT) %>% stable()

st_new(stdata()) %>% st_rename(.list = list(WT = "weight")) %>% stable()
```

# Details
This was discussed quite a bit.  We understand that there is a convenience angle for making `.list` to be `old = new` and a potential for confusion because the specification is reversed.  I decided against using `.dots` as the argument name precisely because I didn't want users to think it would be the same polarity as `...`.  But we might add that `.dots` argument expecting it to be `new = old` for convenience if that were needed or in hopes that the contrast with `.list` would bring clarity to the situation.


kylebaron:
@KatherineKayMRG I didn't know what to call this so I'm calling it `.map` for now ... like a mapping between the old name and the new name.  Meh.  If you have other ideas would love to hear them.  

This sort of thing is called `table` in the `pt_cat/cont_long/wide` functions which I'm not crazy about either.  But `table` (or `.table` would be the name that would be most consistent with what we're doing other places.

KatherineKayMRG:
@kylebaron the code looks good but I agree the name isn't quite right. I also tend to forget that the feature is called .table in the other functions. Given that this is an "opt-in" option, could it have a longer, more intuitive name? I see in `pt_cat_long` you use `all_name_span` - could we do something similar here? e.g. `rename-by-list` or `rename-list`... it's not a single word but it's maybe easier to interpret

kylebaron:
@KatherineKayMRG Good points.  How about just `.list`; that would prevent us from having to type `rename` twice?

```r
st_rename(.rename_list = ...
```

versus

```r
st_rename(.list = ...)
```

tests/testthat/test-table-object.R
```diff
@@ -157,6 +157,16 @@ test_that("stobject equivalent rename", {
   expect_identical(x$output, y$output)
 })
 
+test_that("rename by named list", {
+  mt <- mtcars[1:3,]
+  y <- st_new(mt) %>%
+    st_rename(cylinder = cyl, .list = list(hp = "HorsePower")) %>%
```
annacnev, line 163: this is what I mean by "mixed both". I think the `.list` arg makes most sense as an alternative to the `...` so one or the other (and if you do use both,`.list` will be taken)

Also:

```
cylinder = cyl, .list = list(hp = "HorsePower")
```

This may lead to some headaches. In the `.list` you need to do old name = new name.. but in the `...` you are doing new name = old name. 

Changing it to the code I wrote above for `st_rename` would standardize these to old name = new name. And then this test would become:

```
test_that("rename by named list", {
  mt <- mtcars[1:3,]
  y <- st_new(mt) %>%
    st_rename(.list = list(hp = "HorsePower")) %>%
    inspect2()

  x <- st_new(mt) %>%
    st_rename("cyl" = "cylinder") %>%
    inspect2()
  
  # pass both ... and .list
  # should only change what's in .list
  z <- st_new(mt) %>%
    st_rename("cyl" = "cylinder", .list = list(hp = "HorsePower")) %>%
    inspect2()

  tst <- y$cols_new
  tst2 <- x$cols_new
  tst3 <- z$cols_new

  expect_identical(tst[4], "HorsePower")
  expect_identical(tst2[2], "cylinder")
  expect_identical(tst3[2], "cyl")
  expect_identical(tst3[4], "HorsePower")
})
```

Notice: if we go this route,  `...` has to be strings instead of symbols because we are getting rid of the `enquos`

annacnev (changes requested):
Pitching an alternative way to do this- old method works fine but I think this one is more streamlined and could be less confusing for users.

tests/testthat/test-table-object.R
kylebaron: We are using `old-name = new-name` in `.list` because that is more natural for working with the yspec object.

dpastoor (approved):
changes since last review seem reasonable to me, additionally based on offline discussions

Issue #181, opened 2021-01-12
Return invisible after saving or previewing


kylebaron:
# Summary

As a user, I want pmtables to return invisible output after previewing a table or saving it to file.

# Tests
- tests/testthat/test-table-utils.R
  - test-table-utils-stable_save

PR #192, opened 2021-01-27
Return invisible; some refactors for preview

kylebaron:
#193 #181 

# Summary
- generate error if you try to texPreview a long table
- return `stable` or `stable_long` object  invisibly form `stable_save`
- `st2article` now returns input tables as a list / invisibly
-  When a list is passed to `stable_save` we'll map across that and try to save what is in that list; we pass only`.dir` - file names will have to come from the object

### DIFF
--- a/R/preview.R
+++ b/R/preview.R
@@ -102,6 +102,9 @@ st_preview <- function(x,...) { # nocov start
   if(inherits(x, what = c("pmtable", "stobject"))) {
     x <- as_stable(x)
   }
+  if(inherits(x, what = "stable_long")) {
+    stop("cannot preview long tables in the viewer window", call.=FALSE)
+  }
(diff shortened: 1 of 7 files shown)


kylebaron:
@KatherineKayMRG could you look at this too?  I want to get to your ideal workflow.  Was thinking that if `stable_save` returned the table invisibly, then you could pipe it right to the previewer.  I did sort of the same thing with `st2report` which should return a list of tables invisibly.

dpastoor:
LGTM from a technical perspective - @KatherineKayMRG can you also give a thumbs up on behavioral change

KatherineKayMRG:
This change would be great for me. I already use `%T>%` to pipe tables to `stable_save` and then pipe it to a previewer so this will save me having to remember to add the `T` to the appropriate pipe :)

PR #197, opened 2021-02-07
release version; update news

dpastoor:
looks like across has gotten more picky, it now fails with
✖ object 'ID' not found


kylebaron:
@dpastoor I believe the tests are fixed so that they work `across()` dplyr versions.  I didn't do any other housekeeping (e.g. messing with DESCRIPTION)

Issue #203, opened 2021-03-03
Arguments are not captured when coercing pmtable output to stable_long


kylebaron:
# Summary

As a user, I want to be able to convert pmtable output to long table and pass along arguments to stable_long

# Tests
- tests/testthat/test-as_stable.R
  - pass args to stable_long when coercing stobject
  - pass args to stable when coercing stobject
  - pass args to stable_long when coercing pmtable
  - pass args to stable when coercing pmtable

kylebaron:
When you make a pmtable like this and want to coerce to long table

```r
stdata() %>% pt_cont_long(cols = "WT, ALB, SCR") %>% stable_long()
```
You should be able to pass in arguments to `stable_long()` (like a caption)

```r
stdata() %>% pt_cont_long(cols = "WT, ALB, SCR") %>% stable_long(lt_cap_text = "A caption")
```

That isn't working right now after refactoring `stable_long` to be generic. 

The workaround is 
```r
stdata() %>% pt_cont_long(cols = "WT, ALB, SCR") %>% do.call(what = stable_long)
```

PR #204, opened 2021-03-03
Fix/pass longtable

kylebaron:
See #203 

# Summary
Root cause was that we were building a list of valid arguments in the pmtable object that could get passed to `stable_long`; the valid argument list was still pulling formals from `stable_long`, but `stable_long` was made generic and we needed to look at formals for `stable_long.data.frame`. 

I yanked the code that consults the list of valid arguments. 

In addition to addressing the issue with longtable, this PR also adds tests for the following conversions:
  - pmtable to stable
  - pmtable to stable_long
  - stobject to stable
  - stobject to stable_long

For the coersion from stobject, I removed code that checked argument names; they are all passed along to downstream functions.

I updated lots of the documentation for `st_` functions to be more precise about what argument is getting set and where the arguments are getting passed to. 

This also adds several examples to `demo-pipe.Rmd`

Screen Shot 2021-03-04 at 9 05 09 AM

Screen Shot 2021-03-04 at 9 01 03 AM


KatherineKayMRG (approved):

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # pmtables (development version)
 
+- Fix bug where arguments could not be passed along to stable_long
```
KatherineKayMRG, line 3: do you need to add the font size update here too?

R/table-object.R
```diff
@@ -136,8 +131,9 @@ st_panel <- function(x,...) {
 
 #' Add note information to st object
 #'
-#' See the `notes` and `note_config` arguments to [stable()]. The function can
-#' be called multiple times and will accumulate `notes` data.
+#' See the `notes` and `note_config` arguments passes to [stable()] and then to
```
KatherineKayMRG, line 134:
```suggestion
#' See the `notes` and `note_config` arguments passed to [stable()] and then to
```

NEWS.md
kylebaron: @barrettk is reviewing that PR for me; will get that into the news too.

inst/demo-pipe.Rmd
```diff
@@ -424,3 +424,39 @@ sl %>% st_new() %>% st_sizes(row = 2  ) %>% stable() %>% st_asis()
...
+# Render to long table
+
+```{r}
+longd <- bind_rows(data, data, data, data)
+
+longd %>% 
+  st_new() %>% 
+  st_files(r = "demo-pipe.Rmd") %>%
+  st_notes("This example demonstrates how to make long table") %>%
+  st_args(lt_cap_text = "Adding a caption to the table") %>%
```
KatherineKayMRG, line 439: How about adding the label here too? That seems to be something people often ask about when making long tables

NEWS.md
kylebaron: I added this but didn't render the document; would cause a conflict with the PR that Kyle is reviewing.

inst/demo-pipe.Rmd
kylebaron: I added this but didn't render the document; would cause a conflict with the PR that Kyle is reviewing.

inst/demo-pipe.Rmd
```diff
@@ -424,3 +424,39 @@ sl %>% st_new() %>% st_sizes(row = 2  ) %>% stable() %>% st_asis()
...
+```
+
+\clearpage 
+
+# Pass any argument to stable
+
+Rather than passing dots to `stable()` or `stable_long()`, you can 
+use `st_args()`.  But note that this is destructive to previous work; it assumes 
+you are completely specifying the contents for a given argument.  See this 
+example,
```
KatherineKayMRG, line 452: I appreciate the functionality but is this something we  expect people to need often? I'm struggling to think of when this might be needed... would it be worth suggesting that we recommending using one or the other, as you did above with 
>It is recommended to use either [st_units()] or [st_args()] but not both.

inst/demo-pipe.Rmd
KatherineKayMRG: I assume I don't need to wait for that to appear to approve the PR?

inst/demo-pipe.Rmd
kylebaron: @KatherineKayMRG I think if people want to keep with the pipe approach, they can just keep issuing calling `st_args()` with new arguments.  We can't prevent people from passing stuff directly to `stable()` and I wanted to be transparent about what `st_args()` was doing.

KatherineKayMRG: Oh I see, that makes sense

KatherineKayMRG (approved):
Looks good. Thanks @kylebaron

inst/demo-pipe.Rmd
kylebaron: I pushed some text saying you shouldn't need to do both.

KatherineKayMRG: thanks :)

Issue #206, opened 2021-03-04
Font size not changed for longtable


kylebaron:
# Summary

As a user, I want to control the base font size for longtable

# Tests
- tests/testthat/test-longtable.R
  - set font size in longtable

PR #207, opened 2021-03-04
Fix/longtable fontsize

kylebaron:
Issue #206

Merging this into the other long-table branch that @KatherineKayMRG is reviewing. 

- Don't hard-code font size
- Take font size from stable object
- Add test
- Add example to longtable Rmd


barrettk (approved):
Looks good to me!

barrettk (approved):

PR #208, opened 2021-03-05
Release/0.3.3

kylebaron:
# Summary 

- Fix bug where arguments could not be passed along to `stable_long()` 
  when coercing pmtable output (#203)

- Fix bug where user-specified font size was not getting propagated to 
  long table output with `stable_long()` (#204)


dpastoor (changes requested):
lets bump and 🚢 !

PR #226, opened 2021-05-12
Fix/new names chr

kylebaron:
# Summary 
- @barrettk realized that `new_names.character()` doesn't normalize names unless all the items have alternates
- This PR allows items that are either renamed or not (see the test)


barrettk (approved):
Looks good to me as long as the value changes were expected

kylebaron:
@barrettk I think the values change are a side effect from some work that I did in another branch; we will have to re-assess this when we take up that other branch.

PR #227, opened 2021-05-12
Feature/demographics

barrettk:
# Summary 
- summarize categorical and continuous data in a single table

## Vignette / Examples

Update: 17 Aug 2021
[test-demographics.pdf](https://github.com/metrumresearchgroup/pmtables/files/7004399/test-demographics.pdf)

Update: 10 Aug 2021
[test-demographics.pdf](https://github.com/metrumresearchgroup/pmtables/files/6962030/test-demographics.pdf)

Update: 9 Aug 2021
[test-demographics.pdf](https://github.com/metrumresearchgroup/pmtables/files/6955855/test-demographics.pdf)

Initial version:
[test-demographics.pdf](https://github.com/metrumresearchgroup/pmtables/files/6940112/test-demographics.pdf)

## Required arguments
- `data` - a data frame to summarize
- `cols_cont` - a vector of names (possibly renamed) of continuous data columns
- `cols_cat` - a vector of names (possibly renamed) of categorical data columns

## Optional
- `span` - the column name containing spanner information (possibly renamed)
- `units` - a named list of units ... `list(WT = "(kg)")`
- `stat_name`  - a label for the left hand side of the table listing summary stats
- `stat_width` - the width (in cm) of the `Statistic` column
- `all_name` - the column name for the `All data` summary
- `paneled` - should the table be paneled?  Or wider format?
- `fun` - a function to summarize continuous data
- `notes` - replace the default notes

## Return value
- An object with class `pmtable`; this is a list and the summarised data are in `data` slot

## Summary function
- `...` is a required argument; there will be an error if the user specifies a custom summary function and doesn't have `...` in the signature
- arg 1: `value` ... the data to summarize
- arg 2: `name` ... the name of the column that is being summarized
- arg 3: `...` other arguments passed to the format function
- `fmt` the default format function
- `digits` passed to `fmt`
- `maxex` passed to `fmt`

Note that the default `fmt` function is `pmtables::sig()`; the signature for `sig()` as well as `digit1()` have been expanded to include `...` so that can absorb any unused argument; I'm also exporting `rnd()` which is just `round()` with `...`. 

## Additional changes
Along the way, we made these changes
- Refactor `new_names.character` to handle mix of named and non-named items
- Added utility function `ensure_parens` which will look for items in a vector or list and wrap in parens if they are not detected
- Added example to demo-docs and re-built


R/demographics-table.R
```diff
@@ -0,0 +1,80 @@
...
+
+  assert_that(is.data.frame(data))
+
+  cols_cont <- new_names(cols_cont)
+  cols_cat <- new_names(cols_cat)
+
+  span <- new_names(span)
+  span_lvl <- levels(fct_inorder(data[[unname(span)]]))
+
+  cont_table <- pt_cont_long(
```
kylebaron, line 37: @barrettk I think we do want to stay with the previous code since that would allow users to pass a custom function.  Any summaries that we do should be agnostic to column names etc.

R/demographics-table.R
```diff
@@ -0,0 +1,80 @@
...
+  # Change formatting
+  cont_table <- unite(separate(cont_table$data,`Min / Max`,c("min","max"),sep=" / "), "min-max", min:max,sep=" - ")
+  cont_table <- unite(cont_table,'mean (sd)',c("Mean","SD"), sep = " (")
+  cont_table <- mutate(cont_table,'mean (sd)'= paste0(`mean (sd)`,")"))
+  cont_table <- dplyr::select(cont_table,-c(Median,n))
+
+  # rearrange
+  cont_table <- pivot_longer(cont_table, cols = c("mean (sd)", "min-max"))
+  cont_table <- pivot_wider(cont_table, names_from = names(span), values_from = "value")
+  cont_table <- dplyr::select(cont_table,-c(`All data`))
```
kylebaron, line 51: No need to prefix with `dplyr::`; we control the namespace.

R/demographics-table.R
```diff
@@ -0,0 +1,80 @@
...
+#'
+#' @examples
+#'
+#' @seealso [pt_cont_long()], [pt_cat_long()]
+#'
+#' @export
+pt_demographics <- function(data, cols_cont, cols_cat, span, units = NULL,
+                            stat_name = "Statistic", stat_width = 2) {
+
+  assert_that(is.data.frame(data))
```
kylebaron, line 29: Let's check that `cols_cont` and `cols_cat` exist in the data frame.

kylebaron: Look in `check.R` ... there are functions for checking discrete columns and continuous columns.

R/demographics-table.R
```diff
@@ -0,0 +1,80 @@
...
+#'
+#' @return
+#' An object of class `pmtable`
+#'
+#' @examples
+#'
+#' @seealso [pt_cont_long()], [pt_cat_long()]
+#'
+#' @export
+pt_demographics <- function(data, cols_cont, cols_cat, span, units = NULL,
```
kylebaron, line 26: Let's let the user pass a function; rename that `smr()` function to something in line with what we do for say `pt_cont_long` and use it as the default.

barrettk: @kylebaron Oops just saw this. Ill make the change back to the original code then

R/demographics-table.R
```diff
@@ -0,0 +1,173 @@
...
+    all_cols <- as.list(c(cols_cont,cols_cat))
+    names(units) <- names(all_cols[match(names(units),all_cols)])
+
+    unit_names <- lapply(seq(units), function(i){
+      ifelse(units[[i]]!="",paste0(names(units[i]), " (",units[[i]],")"),names(units[i]))
+    })
+    names(unit_names) <- names(units)
+
+    for(i in 1:length(unit_names)){
+      table_data <- mutate(table_data, name = ifelse(name==names(unit_names[i]), unit_names[[i]], .data[["name"]]))
```
kylebaron, line 150:
```suggestion
      table_data <- mutate(table_data, name = ifelse(.data[["name"]]==names(unit_names[i]), unit_names[[i]], .data[["name"]]))
```

R/demographics-table.R
```diff
@@ -0,0 +1,173 @@
...
+#' @param stat_width distance (in cm) between the statistic and summarized columns
+#'
+#' @return
+#' An object of class `pmtable`
+#'
+#' @examples
+#'
+#' # Example summary function
+#' new_fun <- function(value = seq(1,5)) {
+#'  tibble(
```
kylebaron, line 63:
```suggestion
#'  tibble::tibble(
```

kylebaron:
2021-08-05: invited internal review.

R/demographics-table.R
```diff
@@ -0,0 +1,326 @@
...
+#' pmtables:::dem_cont_fun(value = seq(1,7), digits = 2)
+#'
+#' @keywords internal
+dem_cont_fun <- function(value = seq(1,5), digits = 3) {
+  Mean  <- sig(mean(value,  na.rm = TRUE), digits = digits)
+  Sd    <- sig(sd(value,    na.rm = TRUE), digits = digits)
+  Range <- sig(range(value, na.rm = TRUE), digits = digits)
+  tibble(
+    `mean (sd)` = paste0(Mean, " (", Sd, ")"),
+    `min-max` = paste0(Range, collapse = " - "),
```
timwaterhouse, line 48: Per the style guide, could these both be changed to en-dashes to indicate range, to distinguish from hyphen?
```
    `min--max` = paste0(Range, collapse = " -- "),
```

KatherineKayMRG: The current pmtables functions actually show `Min / Max` should we use that in this function too so all tables are consistent? 
Screen Shot 2021-08-09 at 9 57 24 AM

Is it possible to make all of these outputs consistent with the other tables? So `Mean (SD)`, `Min / Max` and `Non-missing`

KatherineKayMRG: Two more thoughts on the non-missing row

1. Grammatically it should read `Nonmissing` (without the hyphen)
2. Is it too late to swap this to show `Missing` as we do in other tables? So in this table, instead of 65, 92 and 157 nonmissing in the Weight row, could it read 1, 2, 3 Missing? 

Screen Shot 2021-08-09 at 10 21 00 AM

R/demographics-table.R
```diff
@@ -0,0 +1,326 @@
...
+  cont_table <- mutate(
+    cont_table,
+    name = factor(.data[["name"]], levels = unname(cols))
+  )
+  cont_table <- arrange(cont_table, .data[["name"]])
+  cont_table <- mutate(cont_table, name = names(cols)[.data[["name"]]])
+  cont_table
+}
+
+pt_demographics_notes <- function() {
```
KatherineKayMRG, line 318: Is `n: number of records summarized` included when it appears in the column header?

R/demographics-table.R
kylebaron: Will change to `Min / Max` per @KatherineKayMRG suggestion.

kylebaron: Also will report `Missing` per @KatherineKayMRG suggestion.

R/demographics-table.R
kylebaron: Yes; I added a note there.

kylebaron:
I made the changes suggested by @timwaterhouse and @KatherineKayMRG ; It's not too late to change anything ... this is in fact why we're doing the review - to make it look exactly the way we want and to do it befire the release. Please take a look at the updates and make sure it is what you wanted. Also please suggest other changes you'd like to see.

KatherineKayMRG:
More of a questions... For tables like this where span isn't provided, is it still possible to add the total n under "All data"?

Screen Shot 2021-08-09 at 12 19 26 PM

The examples seem to show n appearing only on tables with span, just wondering if this is a coincidence. Ideally, I'd like to add n to this kind of table too

kylebaron:
@KatherineKayMRG Please refer to the updated example doc. Thanks.

KatherineKayMRG:
@kylebaron  Absolutely - can you point me to it?

kylebaron:
Scroll to the top.

KatherineKayMRG:
Actually, I see it now (needed to refresh the page).

KatherineKayMRG:
I know this is getting really picky now but could the continuous notes be ordered in the same way as they are in the current table functions? That way if someone has multiple tables in a report their footnotes will be consistent?

Screen Shot 2021-08-09 at 12 37 34 PM

kylebaron:
Yep.

kylebaron:
Let me know what else to change.

KatherineKayMRG:
No other changes are jumping out at me regarding cosmetics of the tables. I was just looking closer at the documentation... it would be nice to see an example of how people could grab their units from the yspec in section 1.3:
Screen Shot 2021-08-09 at 12 48 51 PM

Also, I typically use the `table` option to both rename covariates and add units (e.g. `covlab = ys_get_short_unit(spec, parens = TRUE)` and `table=covlab` in the function), could we get an example of that here too?

kylebaron:
@KatherineKayMRG I updated the vignette (scroll to top) with table example for both continuous and categorical. 

I think this is a good area to pursue ... what else can we do with `pt_cont_long()` tables that is reasonable to do in the demographics table but isn't currently implemented?

Thanks for reviewing.

KatherineKayMRG:
How about an example the uses that flags in the yspec? I often add continuous and categorical flags in my yspec

```
edaCont: [WT, AGE]
edaCat: [SEX, RACE]
```
and then use the with `ys_filter`...
```
pt_demographics(
    pmt_first,
    cols_cont = names(ys_filter(spec, edaCont)),
    cols_cat = names(ys_filter(spec, edaCat)),
    span = c(Study = "STUDYf"),
    table = tab
)
```

KatherineKayMRG:
@timwaterhouse - @kylebaron and I were discussing having multiple built in options for this function. What do you think of the suggestions in the screenshot? 

alternative

timwaterhouse:
@KatherineKayMRG The idea is that these built-in options will be available to choose from, with one as default? And the ability to override with manual stats? That seems reasonable to me. I like the suggestions here.

kylebaron:
@timwaterhouse We were providing one default summary function with the option for users to create their own. This would still have the default, but we'd provide alternate summaries that users could pick from (`summary_2()` rather than `summary_1()` and still retain the ability to pass their own function to get completely customized).

KatherineKayMRG:
@timwaterhouse exactly. So by default, the user would get the option on the left of the screenshot but, as Kyle said, they could either choose to display the middle or right options... or they can still write their own function. Are there any other built-in summary options you think would be useful?

curtisKJ:
An option with 95% CI as an alternate to the min/max would be useful.

kylebaron:
Thanks @curtisKJ; assuming you want 5th and 95th percentiles rather than CI;  @KatherineKayMRG could we swap out Min / Max for that in one of the options? Maybe add it to option 3?

kylebaron:
Sure; can you suggest what the label should be for this?

curtisKJ:
90% interval

kylebaron:
Thx.

kylebaron:
Please hold on reviewing until WIP tag is off the title.

barrettk (commented):
Cant request changes since I initiated the pull request, but I noticed that the build is outdated (pmtables_0.3.3.9006 vs pmtables_0.3.3.9000), which I believe could affect validation scripts. `dem_cont_fun` changed between iterations.

PR #249, opened 2021-09-01
Feature/demographics

kylebaron:
# History
See #227; Kyle Baron opened up the current PR so that @barrettk and others could comment and review. The summary and notes are copied here as well. 

# Summary 
- summarize categorical and continuous data in a single table

## Vignette / Examples

Update: 17 Aug 2021
[test-demographics.pdf](https://github.com/metrumresearchgroup/pmtables/files/7004399/test-demographics.pdf)

Update: 10 Aug 2021
[test-demographics.pdf](https://github.com/metrumresearchgroup/pmtables/files/6962030/test-demographics.pdf)

Update: 9 Aug 2021
[test-demographics.pdf](https://github.com/metrumresearchgroup/pmtables/files/6955855/test-demographics.pdf)

Initial version:
[test-demographics.pdf](https://github.com/metrumresearchgroup/pmtables/files/6940112/test-demographics.pdf)

## Required arguments
- `data` - a data frame to summarize
- `cols_cont` - a vector of names (possibly renamed) of continuous data columns
- `cols_cat` - a vector of names (possibly renamed) of categorical data columns

## Optional
- `span` - the column name containing spanner information (possibly renamed)
- `units` - a named list of units ... `list(WT = "(kg)")`
- `stat_name`  - a label for the left hand side of the table listing summary stats
- `stat_width` - the width (in cm) of the `Statistic` column
- `all_name` - the column name for the `All data` summary
- `paneled` - should the table be paneled?  Or wider format?
- `fun` - a function to summarize continuous data
- `notes` - replace the default notes

## Return value
- An object with class `pmtable`; this is a list and the summarised data are in `data` slot

## Summary function
- `...` is a required argument; there will be an error if the user specifies a custom summary function and doesn't have `...` in the signature
- arg 1: `value` ... the data to summarize
- arg 2: `name` ... the name of the column that is being summarized
- arg 3: `...` other arguments passed to the format function
- `fmt` the default format function
- `digits` passed to `fmt`
- `maxex` passed to `fmt`

Note that the default `fmt` function is `pmtables::sig()`; the signature for `sig()` as well as `digit1()` have been expanded to include `...` so that can absorb any unused argument; I'm also exporting `rnd()` which is just `round()` with `...`. 

## Additional changes
Along the way, we made these changes
- Refactor `new_names.character` to handle mix of named and non-named items
- Added utility function `ensure_parens` which will look for items in a vector or list and wrap in parens if they are not detected
- Added example to demo-docs and re-built


DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: pmtables
 Type: Package
 Title: Tables for Pharmacometrics
-Version: 0.3.3.9000
+Version: 0.3.3.9006
```
kylebaron, line 4: @barrettk What should I change for "build"?

R/demographics-table.R
```diff
@@ -0,0 +1,363 @@
...
+#' call the default summary function but with different formatting parameters.
+#'
+#' The summary function should return a data frame, with summary statistics
+#' in the columns.
+#'
+#' @examples
+#' pmtables:::dem_cont_fun(value = seq(1,7), digits = 2)
+#'
+#' @keywords internal
+dem_cont_fun <- function(value = seq(1,5), name = "",  ..., fmt = sig,
```
kylebaron, line 64: @barrettk The summary function was updated in subsequent development cycles. Is there some change that we should see?

barrettk: I think you just have to run `devtools::build()` and put the new tarball in the directory. Without running `load_all()`, there are a few functions that are outdated. I dont think that will matter once you make a release, but may be worth updating for validation purposes
```
> library(pmtables)
> dem_cont_fun
Error: object 'dem_cont_fun' not found
> pmtables:::dem_cont_fun
function(value = seq(1,5), digits = 3) {
  tibble(
    `mean (sd)` = paste0(sig(mean(value, na.rm = TRUE), digits), " (", sig(sd(value,na.rm=TRUE),digits), ")"),
    `min-max` = paste0(sig(range(value,na.rm=TRUE), digits), collapse = " - ")
  )
}
```

kylebaron: @barrettk There is no tarball in the directory. Did you build and install the package on the branch?

kylebaron: You can do this on the command line with `make install`

kylebaron: Or `R CMD install`

barrettk (approved):
Ran all tests/checks and went through the new code/examples. Looks good to me

R/demographics-table.R
barrettk: Yeah I did that too and there wasn't any issues. I assumed the tarball may have been used for validation purposes because I noticed a script that did so in the mrgsolve package. Everything else looked good though

Issue #236, opened 2021-08-04
Call st_asis on pmtable object


kylebaron:
# Summary

As a user, I want to be able to pipe a `pmtable` object to `st_asis()`

# Tests
- tests/testthat/test-preview.R
  - all st_asis on a pmtable object

Issue #232, opened 2021-07-29
Escape names in table list for st2article


kylebaron:
# Summary

As a user, I want pmtables to escape names in the list of tables to render with st2report

# Tests
- tests/testthat/test-preview.R
  - st2report - list names are escaped

Issue #231, opened 2021-07-08
Reverse order on span_split


kylebaron:
# Summary

As a user, I want to be able to take the span title from the left or right split when using `colsplit()`

# Tests
- tests/testthat/test-span.R
  - span split with reversed title / col

Issue #224, opened 2021-05-11
Add newline after st_asis


kylebaron:
# Summary

As a user, I want pmtables to add a newline after st_asis() is called

kylebaron:
See #219

PR #250, opened 2021-09-01
Release/0.4.0

kylebaron:
# NEWS

- Add `cols_omit` option to omit column header data  (#213)

- Add `pt_demographics` function to generate a new table of both continuous and
  categorical data (#186, #249)

- Add `title_side` argument to `colsplit()` so that the title can be taken from
  left or right side of split (#231)

- Add `hline` argument to `rowgroup()` constructor to make the horizontal line
  above the panel data optional (#215)

- Refactor `pt_data_inventory()` to calculate percent BLQ using denomninator
  that is the sum of the number of observations BLQ and non-BLQ / non-missing
  (#221, #222)

- `st_asis()` gains a method for `pmtable` objects (#236)

- List names are now escaped when passing a list of tables to `st2report()` 
  and friends (#232)

- Add newline after printing table text using `st_asis()` (#224)


dpastoor (approved):
👍

PR #255, opened 2022-01-14
Refactor/bq footer

kylebaron:
# Summary 

- make BLQ / BQL consistent in table header and notes

Issue #254, opened 2022-01-13
pt_data_inventory to update colname/footer when I use BLQ


KatherineKayMRG:
# Summary

As a user, I want pt_data_inventory to change my colname/footer description for BLQ if I use `bq_col = "BLQ"`

# Tests
- tests/testthat/test-inventory-table.R
  - handle BQL and BLQ inventory table

kylebaron:
How about this?

``` r
library(pmtables)


bq_col <- "BQL"
pt_data_inventory_notes(bq_col)
#> [1] "SUBJ: subjects"                      
#> [2] "BQL: below quantitation limit"       
#> [3] "MISS: missing observations (not BQL)"
#> [4] "OBS: observations"
bq_col <- "BLQ"
pt_data_inventory_notes(bq_col)
#> [1] "SUBJ: subjects"                      
#> [2] "BLQ: below limit of quantitation"    
#> [3] "MISS: missing observations (not BQL)"
#> [4] "OBS: observations"
```

<sup>Created on 2022-01-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

kylebaron:
Hang on; need to fix `MISS` then.

kylebaron:
``` r
library(pmtables)


bq_col <- "BQL"
pt_data_inventory_notes(bq_col)
#> [1] "SUBJ: subjects"                   "BQL: below quantitation limit"   
#> [3] "MISS: missing observations (BQL)" "OBS: observations"
bq_col <- "BLQ"
pt_data_inventory_notes(bq_col)
#> [1] "SUBJ: subjects"                   "BLQ: below limit of quantitation"
#> [3] "MISS: missing observations (BLQ)" "OBS: observations"
```

<sup>Created on 2022-01-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

kylebaron:
And then handling if we just don't want BLQ

``` r
library(pmtables)

bq_col <- "BQL"
pt_data_inventory_notes(bq_col)
#> [1] "SUBJ: subjects"                   "BQL: below quantitation limit"   
#> [3] "MISS: missing observations (BQL)" "OBS: observations"
bq_col <- "BLQ"
pt_data_inventory_notes(bq_col)
#> [1] "SUBJ: subjects"                   "BLQ: below limit of quantitation"
#> [3] "MISS: missing observations (BLQ)" "OBS: observations"
[...]

Issue #253, opened 2021-12-17
Add * to panel title rows in longtable


kylebaron:
# Summary

As a user, I want pmtables to keep panel title rows together with the subsequent row.

# Tests

- tests/testthat/test-panel.R
  - nopagebreak for panels in longtable

PR #252, opened 2021-12-17
Add jut and nopagebreak as panel options

kylebaron:
# Summary
- add `jut` argument to `rowpanel()` to shift non-panel rows; units are `ex`
- add `nopagebreak` argument to `rowpanel()`; default is `TRUE` and results in `\\*` getting added to panel rows to 
when rendering `longtable` prevent page breaks (in `longtable` environment only)
- panel rows are marked by a commented tag, stored in the `.internal` environment
- Filled in documentation for `longtable` and added **simple** example

# Example - jut
Screen Shot 2021-12-21 at 12 15 30 PM

# Example - nopagebreak

```
> stable_long(data1, panel = "STUDY")
....                                                                 
[11] "\\hline"                                                                        
[12] "DOSE & FORM & N & WT & CRCL & AGE & ALB & SCR \\\\"                             
[13] "\\hline"                                                                        
[14] "\\endfirsthead"                                                                 
[15] "\\hline"                                                                        
[16] "DOSE & FORM & N & WT & CRCL & AGE & ALB & SCR \\\\"                             
[17] "\\hline"                                                                        
[18] "\\endhead"                                                                      
[19] "\\hline"                                                                        
[20] "\\multicolumn{8}{l}{\\textbf{12-DEMO-001}}\\\\*%--pmtables-insert-panel"        
[21] "100 mg & tablet & 80 & 71.4 & 104 & 33.7 & 4.20 & 1.06 \\\\"                    
[22] "150 mg & capsule & 16 & 89.4 & 122 & 24.4 & 4.63 & 1.12 \\\\"                   
[23] "150 mg & tablet & 48 & 81.7 & 104 & 34.4 & 3.83 & 0.910 \\\\"                   
[24] "150 mg & troche & 16 & 94.0 & 93.2 & 27.4 & 4.94 & 1.25 \\\\"                   
[25] "200 mg & tablet & 64 & 67.9 & 100 & 27.5 & 4.25 & 1.10 \\\\"                    
[26] "200 mg & troche & 16 & 76.6 & 99.2 & 22.8 & 4.54 & 1.15 \\\\"                   
[27] "\\hline \\multicolumn{8}{l}{\\textbf{12-DEMO-002}}\\\\*%--pmtables-insert-panel"
[28] "100 mg & capsule & 36 & 61.3 & 113 & 38.3 & 4.04 & 1.28 \\\\"                   
[...]

PR #258, opened 2022-03-08
Release/0.4.1

kylebaron:
- [ ] version number in DESCRIPTION
- [ ] version number in NEWS

# Summary 

- `colgroup()` (and `st_span()`) gains an `align` argument to position
  the spanner title on left or right in addition to the center (default)
  #260, #261.

- `rowpanel()` (and `st_panel()`) gains `jut`argument to push non-panel
  table contents to the right relative to the panel header row so that
  contents under the panel header are indented #251, #253. 

- Panel header rows are now modified so that the header row stays
  with the first non-header row for longtable output #252, #253.

- Consistent `BQL` / `BLQ` handling for column titles and table 
  notes for `pt_data_inventory()` #254, #255.


seth127 (approved):
Looks good to me.

Only comment would be whether you want to add examples of any of these new features to the "book" but not necessary to do that before the release.

PR #261, opened 2022-03-13
Feature/align span

kylebaron:
Pass along alignment when rendering spanner.  I wouldn't use it in this case, but have been seeing some geometries that look weird but improved when the columns and span title are left aligned.  

Default is `c`; also accepts `r`

```r
library(pmtables)

data <- stdata()

st_new(data) %>%
  st_span("On the left", WT:SCR, align = 'l') %>% 
  stable() %>% 
  st2report()

```

Screen Shot 2022-03-14 at 12 50 54 AM


KatherineKayMRG (approved):
Looks like a nice feature. I agree it would be helpful to have this option.

PR #268, opened 2022-03-25
Features - complete + cat denominator

kylebaron:
# Summary 

## Part 1

- `pt_cat_wide`  completes combinations of `by` and `panel` 
- This was by user request and we do not `complete` by default
- There is a test in place that checks that factor level order is preserved in the output
- There is no partial completion; you will get completion of any / every combination as implemented by `dplyr::complete`

## Part 2
- Adds `denom` argument to the following tables
  - `pt_demographics()`
  - `pt_cat_wide()`
  - `pt_cat_long()`
- Choices are
  - `group` - use the group record count to calculate percents
  - `total` - use the total record count to calculate percents
- The default remains the same (`group`) but opt into `total`
- `pt_demographics()` uses `pt_cat_long()` so test for `denom` appeals to that output which is tested more explicitly

## Vignettes

These vignettes demonstrate the new functionality

[cat-denom-refactor.pdf](https://github.com/metrumresearchgroup/pmtables/files/8375312/cat-denom-refactor.pdf)
[complete.pdf](https://github.com/metrumresearchgroup/pmtables/files/8375313/complete.pdf)


barrettk (approved):
Looks good. Ran tests and walked over code. Didnt notice any spelling mistakes. Did you run that command that looks for spelling mistakes already?

kylebaron:
Thanks, @barrettk, for reviewing; I did the spell check; looks ok.

PR #274, opened 2022-04-20
Feature/pmtable st new

kylebaron:
# Summary 

Two main objectives with this PR

1. Allow a `pmtable` object to enter the table pipe line via `st_new()`
2. Enhance ability to manipulate the table `notes` in the pipeline; this is related to the first object in that we are in need of tweaking notes that are automatically generated by the `pmtable` functions

## Send `pmtable` into `st_new()`

- `st_new()` is now generic with methods for `data.frame` and `pmtable`
- The `pmtable` method takes the pmtable object (already a list), and just makes a copy of everything in a new `stobject
- We are tracking "valid" fields in the pmtable object for safety

## Notes manipulation

- `st_notes_detach()` - make the table notes appear in a minipage with a certain width; other things are customizable here, but this is the biggie
- `st_notes_rm()` - remove the `notes` vector in the object; again, mainly for `pmtable` objects where we might want to nuke the automatic notes and then start over
- `st_notes_app()` - append a note to the end of the previous line
  - we can also do this with `append` argument to `st_notes()`; I wanted a standalone function 
- `st_notes_str()`  - (to string) - collapse notes into a single string 
  - we can also do this with `to_string` argument to `st_notes()`
- `st_notes_conf()` - is an alias to `st_noteconf()` which fits better with the naming scheme 
- `st_notes_sub()` - select one _line_ in the table notes by a regular expression or integer location and replace that _entire_ line; this is _simple_ editing capability


KatherineKayMRG (approved):
Looks good

PR #277, opened 2022-04-30
Table as Image

kylebaron:
# Summary

This PR implements functionality to render a `stable` object in image format to be saved on disk and / or to be included inline in an `html` Rmarkdown document. 

- `st_aspdf()` - write table to a standalone tex template and render with `pdflatex`
- `st_aspng()` - write table to a standalone tex template, render with `latex` and convert with `dvipng`
- `st_as_image()` - a one liner that renders the table as pdf and then reads it back in using `magick:::read_image_pdf()`; this is generic with methods for 
  - `stable`
  - `pmtable`
  - `stobject`
- `st_show_image()` - exposed worker that displays image object based on current context
  - If knitting, then the image width is resized to a percentage of the available space
  - If not knitting, then the image width is resized to a percentage of the graphics device
- In case the image fails to render, you will still get back an image, but the image will include text notifying you that the build failed
- When building, you can set `textwidth`
- When building, you can set `border` dimensions
- When building, you specify the `build` directory
- When building, you can specify the `stem` for output files
- When building via `png` the resolution can be passed through as `dpi`
- Unexposed option (yes ... I said it) to revert to older version of longtable; the newest version doesn't work well with the setup we have; but it builds fine with a previous version; going to go with this until we see what happens in the wild; this option is called `pmtables.image.ltversion` and it should give a numeric version number 

# Added to suggests
- `magick`
- `pdftools`

# System dependencies
- `pdflatex`
- `latex`
- `dvipng`

# Tests

See `tests/testthat/image/as-image-test.Rmd`

# Tests


KatherineKayMRG:
Looks like a great new feature and I like that you get back some kind of image if the table fails to build

Just a thought - do you want to follow similar naming conventions to existing preview functions? E.g.

- st_aspdf() - st2pdf
- st_aspng() - st2png
- st_as_image() - st2image

or do you think this will lead to confusion?

KatherineKayMRG (approved):
I didn't dig through the code on this one (I think Seth's better for that) but the behaviour makes sense to me.

seth127 (commented):
Code looks good and functionality looks good. My only questions were about the testing and validation. If you have a method in mind for this stuff then go ahead and merge. The following questions are more for my education:

I see the `as-image-test.Rmd`. I've never seen this pattern before. Does this get run as part of the test suite? It doesn't look like it from me running `devtools::test()` but maybe I missed it. If not, I assume this is just for manual testing? I ran it manually and all the outputs look good to me.

It looks like no unit tests for this. Is that because its tough to test these functions? It seems like you could do all the stuff in the `.Rmd` file in a regular test file and just check that the image files get written (presumably into a temp folder that you'd clean up).

I think you want to add validation stories for this [here](https://github.com/metrumresearchgroup/pmtables/blob/develop/inst/validation/pmtables-stories.yaml) but if we're not gonna have unit tests then they aren't worth much. Wondering what your thoughts are on that.

Thanks. Great stuff.

kylebaron:
@seth127  - 

Right; the Rmd is the (manual) test; I tested and attached to the PR for the reviewer to look at or run on their own

I'd be open to ideas for the right way to handle this. 

Kyle

kylebaron:
Re-factor test Rmd file to include tests of `.tex` intermediates.

[as-image-test.zip](https://github.com/metrumresearchgroup/pmtables/files/8664955/as-image-test.zip)

kyleam:
> Re-factor test Rmd file to include tests of .tex intermediates.

Thanks.  What do you think about moving these to a normal test file so that they can be executed by the CI?  Something like https://github.com/metrumresearchgroup/pmtables/compare/feature/as-image...km/feature/as-image-unit

If that direction looks okay to you, of course feel free to bring that commit (e68f383c9ab) into this PR and tweak as you see fit.  (I didn't add any test IDs.)

PR #280, opened 2022-05-11
as_image: move unit tests to test file

kylebaron:
Put the unit tests in a standard test file rather than the Rmd so that
the tests are executed by the CI.  The price is some duplication
between the Rmd and test file.


kyleam (approved):
Ah, with <https://github.com/metrumresearchgroup/pmtables/pull/277#issuecomment-1123143737>, I was just thinking you'd `git merge --ff-only km/feature/as-image-unit` if you wanted to bring in the commit, but this works for me too (and the end result is the same).

Thanks

kylebaron:
Ok; thanks; I had never done that before.  Got it in there now.

PR #286, opened 2022-05-29
Add blank lines around hline for detached notes

kylebaron:
# Summary

We put an horizontal rule above table notes in this "detached" format. When previewing the table on Metworx (only) with `st_as_image()`, the spacing around the hrule isn't correct. This is solved by adding some blank lines above and below the hrule.  I don't know what is different about the build process on Metworx versus my macbook. But this is a simple fix that has no consequence for tables that show up when rendering a table in a report. 

This pull request adds blank lines around the hrule when creating detached notes. 

We also switch to vspace rather than vskip.

There is also a tweak to the validation script; casualty around keeping this in the code and not someplace else. Keep it and we'll straighten it out later on. 

There are no tests or user stories for this PR; a visual confirmation is provided.


kylebaron:
Screen Shot 2022-07-28 at 12 26 40 PM

kylebaron:
Confirmation of behavior in tex article 

Screen Shot 2022-07-28 at 12 32 47 PM

kyleam (commented):
Looks like automated tests are failing: https://github-drone.metrumrg.com/metrumresearchgroup/pmtables/1475/1/3

Part of it looks like a`vskip -> vspace` needed here (not sure if there's another failure too): https://github.com/metrumresearchgroup/pmtables/blob/f0d90689b04be2894016121661e44e5f93b65a88/tests/testthat/test-notes.R#L37-L38

kyleam (approved):
Tests passing with your latest push (not sure why drone marker isn't showing up next to the commits here):

https://github-drone.metrumrg.com/metrumresearchgroup/pmtables/1484

Looks good/safe as far as I can tell.

kylebaron:
Thanks, @kyleam ; made a couple of changes last minute to make this work better with `standalone` and didn't see that failure.

kylebaron:
For documentation: this is what happened when rendering the table to `standalone` on Metworx before making this change. 

Screen Shot 2022-07-28 at 3 44 33 PM

PR #297, opened 2022-07-28
Fix/inventory all data

kylebaron:
# Summary

- Actually use the `all_name` argument (this wasn't used before)
- Adds `summarize_all` (logical) argument to option out of this row if desired
- Adds `all_name_stacked` argument to provide the name when `stacked` is `TRUE`


kyleam (approved):
I haven't poked around this code base much, so things are a bit unfamiliar, but, reading through the code and playing around with `all_name/summarize_all/all_name_stacked`, things work nicely and as advertised as far as I can tell.

kylebaron:
Thanks @kyleam ; I could have put a better summary in here. But yeah it really just let's the user decide about the "All data" summary, similar to all the other tables. And this one has a couple types of those "All data" (depending on how you configure the table).  So overall it not a big change.

PR #298, opened 2022-08-03
st-filter

kylebaron:
# Summary

This is adds another function to a set that lets you call dplyr verbs on the data in a table pipeline. This started out a little gimmicky (why not work on the data before starting the table pipeline?) but legit need now that we can pass `pmtable` objects into the pipeline.  `pmtable` objects are created after running a standardized data summary and formatting. So you don't have the actual table data before starting the pipeline. So this series of functions let's you tweak some decisions made by the standard data summary process. 

- Adds `st_filter` which calls `dplyr::filter` on the data
- This also cleans up the documentation a bit and adds examples


R/table-object.R
```diff
@@ -885,6 +899,16 @@ st_mutate <- function(x, ...) {
   x
 }
 
+dplyr_filter <- dplyr::filter
```
kyleam, line 902: Just out of curiosity: why do this assignment rather than call `dplyr::filter` directly?  The only thing I could come up with was avoiding the `::` lookup overhead for each `st_filter` call, but that seems too minor to be the reason.

kyleam (approved):
LGTM (and the summary explains the context/motivation nicely).

R/table-object.R
kylebaron: Yeah it's not performance but coding style. I still get nervous about `filter()` but I'm seeing other usage and this should find `dplyr::filter()`.  I'll revert this and trust the lookup.

kylebaron: Just realizing that filter is in stats namespace; so using it unqualified after importing from `dplyr` (and not `stats`) should be safe. Thanks for picking this up.

PR #302, opened 2022-12-09
is_str_regex: account for stringr 1.5.0 class change

kyleam:
The class of pattern modifiers has changed with stringr's 2895a6b (Use stringr_ class prefixes for pattern modifiers, 2022-01-18), part of the v1.5.0 release.

---

This popped up in MPN's check failures.   With this patch, checks pass with either stringr 1.5.0 and 1.4.1.

@kylebaron It seemed easy enough to stay backward compatible here, but let me know if you would instead prefer to just bump the stringr version constraint in DESCRIPTION.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -120,7 +120,10 @@ is_regex <- function(x) {
 
 is_str_regex <- function(x) {
   if(!is.character(x)) return(FALSE)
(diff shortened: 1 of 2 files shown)


kylebaron (approved):

kylebaron:
Thanks for fixing this @kyleam; seems easy enough to accommodate both for now.

PR #313, opened 2023-07-07
Better caption handling for preview and tex files

kylebaron:
# Big picture
- Add simple functionality to create captions for tabular and long tables
- Captions may have a short title that will be used in List of Tables
- No explicit support for labels; users can write these into the caption text
- Captions always show up in `st2report()` and `st2article()` outputs with short titles in ToC (if present)
- Captions for long tables are always written to file (caption is part of the table body)
- Captions for tabular tables may optionally be written to file

## Details

- Added `st_caption()` to set caption in stable pipeline
  - This function mainly runs on an `as.caption()` function that can be used for testing or dealing with captions outside of this pipeline setup; main anticipated access point, though, is `st_caption()`
- Captions passed to `st_caption()` (and `as.caption()`) can include a short title in brackets at the beginning of the caption text; this will be parsed and optionally repeated at the start of the main caption text
- `stable()` and `stable_long()` now  save the caption as an attribute on `stable` object
- Saved captions may also have a `short` attribute on them
- Saved captions also have a `write` attribute which indicates whether the caption will be written to file when calling `stable_save()`
- Modified `pt_wrap()` so that it recognizes caption
  - Captions are _always_ written out (if they exist) when rendering tables via `st2report()` or `st2article()`
- Modified `stable_save()` to optionally write captions out when saving tabular tables

# Example
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(pmtables)

a <-
  stdata() %>%
  st_new() %>%
  st_caption(
    "[Short title]. Main caption text. \\label{tab::example}",
    write = TRUE
  ) %>% stable()

stable_save(a, file = "foo.tex", dir = "~/")

readLines("~/foo.tex")[1]
#> [1] "\\caption[Short title]{Short title. Main caption text. \\label{tab::example}}"

b <-
  stdata() %>%
  st_new() %>%
  st_caption(
    "[Final Model: Population simulations for every 4 week dosing in adults].
    Covariates from the adult analysis population were resampled with
    replacement to a population of N=1,000 and used to simulate
    PK for the following regimen: 50 mg IV every 4 weeks.
    Subject-level variability, but not interoccasion or residual error,
    was included in the simulations. Summary statistics for pre-dose
    concentration and average concentration at steady state were calculated
    across all 1,000 resampled patients. \\label{tab:pop-sims-adult}"
  ) %>% stable_long()

c <-
  stdata() %>%
  st_new() %>%
  st_caption(
    "Full Model: Final model parameter estimates\\label{tab:param}",
    short = "Full Model Estimates", short_repeat = FALSE
  ) %>% stable_long()

st2report(list(a,b,c), ntex = 2, output_dir = "~/")
```

[view-st2report.pdf](https://github.com/metrumresearchgroup/pmtables/files/12051794/view-st2report.pdf)


R/preview.R
```diff
@@ -209,6 +209,14 @@ st2article <- function(..., .list = NULL, ntex = 1,  #nocov start
                        dry_run = FALSE, stdout = FALSE, show_pdf = TRUE) {
 
   tables <- c(list(...),.list)
+  if(length(caption==1) && length(tables) > 1) {
```
kylebaron, line 212: This is previous behavior that we'll retain for now.

R/preview.R
```diff
@@ -248,15 +256,23 @@ st2article <- function(..., .list = NULL, ntex = 1,  #nocov start
   temp <- mgluet(temp, .envir = env)
 
   if(is.character(caption)) {
-    wrap_with_caption <- function(text, i) {
-      if(is.null(i)) i <- "<no name given>"
-      short <- paste0("pmtables output preview - ", i)
+    wrap_with_caption <- function(text, i, short, cap) {
```
kylebaron, line 259: In this helper, we'll replace previous behavior with the new caption info.

R/preview.R
```diff
@@ -248,15 +256,23 @@ st2article <- function(..., .list = NULL, ntex = 1,  #nocov start
...
+        if(is.null(short)) short <- caption
+      }
       pt_wrap(
-        text, context = "tex", caption = caption, short = short,
-        con = NULL
+        text, context = "tex", caption = caption, short = short, con = NULL
       )
     }
-    tables <- imap(tables, wrap_with_caption)
+    tables <- Map(
```
kylebaron, line 269: Need to switch this from `imap()` since we're mapping along vectors of shorts and captions.

R/preview.R
```diff
@@ -430,7 +446,9 @@ pt_wrap <- st_wrap
 #' @rdname st_wrap
 #' @export
 st_asis <- function(x, ..., asis = TRUE, con = NULL) {
-  st_wrap(x, ..., asis = asis, con = con)
+  caption <- cap_main(x)
+  short <- cap_short(x)
+  st_wrap(x, ..., asis = asis, con = con, caption = caption, short = short)
 }
```
kylebaron, line 452: This gets captions and short names into the `st_asis()` output.

R/caption.R
```diff
@@ -0,0 +1,110 @@
...
+#' call.
+#'
+#' @param text a chunk of caption text that may or may not need to be parsed.
+#' @param short optional short caption text
+#' @noRd
+parse_caption <- function(text, short = NULL, short_repeat = TRUE, short_sep = "") {
+  text <- trimws(text)
+  parsed_short <- str_extract(text, "^\\[.+?\\]")
+  has_parsed_short <- !is.na(parsed_short)
+  if(is.null(short) && !is.na(parsed_short)) {
```
kyleam, line 13: The previous line already put the second part of this in a variable, so `is.null(short) && !is.na(parsed_short)` → `is.null(short) && has_parsed_short`?

R/preview.R
```diff
@@ -248,15 +258,23 @@ st2article <- function(..., .list = NULL, ntex = 1,  #nocov start
...
 
   if(is.character(caption)) {
-    wrap_with_caption <- function(text, i) {
-      if(is.null(i)) i <- "<no name given>"
-      short <- paste0("pmtables output preview - ", i)
+    wrap_with_caption <- function(text, i, short, caption) {
+      if(has.st_caption(text)) {
+        caption <- cap_main(text)
+        short <- cap_short(text)
+        if(is.null(short)) short <- caption
```
kyleam, line 265: Just to check my understanding: a NULL `short` isn't actually reachable given the upstream code, right?  Line 217-221 ensures that short is assigned to something non-NULL.

R/table-object.R
```diff
@@ -921,6 +922,33 @@ st_edit <- function(x, ...) {
...
 }
 
+#' Set table caption
+#'
+#' @param x an stobject.
+#' @param text caption text to be parsed.
+#' @param ... additional arguments passed to [as.caption()].
+#'
+#' @details
+#' A short title can also be specified by including it in brackets (`[]`) as
```
kyleam, line 932: This "also" is referring, I think, to the fact that you can pass `short` via `...` to `as.caption`.  But I don't expect that to be obvious to a reader given `short` won't appear in st_caption's documentation.  I think it'd be good to explicitly mention the `short` option.

R/caption.R
```diff
@@ -0,0 +1,110 @@
...
+#' when short is repeated. If short not repeated, then ok to just trim.
+#' trimws() is called quite a bit, but important to consider context of the
+#' call.
+#'
+#' @param text a chunk of caption text that may or may not need to be parsed.
+#' @param short optional short caption text
+#' @noRd
+parse_caption <- function(text, short = NULL, short_repeat = TRUE, short_sep = "") {
+  text <- trimws(text)
+  parsed_short <- str_extract(text, "^\\[.+?\\]")
```
kyleam, line 11: So the two limitations of the "[short] ..." syntax I can think of are

 * short text can't contain a new line character
 * short text can't contain a `]`

For the first, there may not be any cases where it would reasonable for short text to contain a new line (and a single plain `\n` won't be rendered differently by LaTeX anyway).

For the second, I guess there could be some cases, but then you can use the `short` argument.

tests/testthat/test-caption.R
```diff
@@ -0,0 +1,217 @@
+library(testthat)
+library(dplyr)
+
+context("test-caption")
```
kyleam, line 4: `context` is deprecated but may still worth adding for consistency with other tests.

kyleam (approved):
Thanks for the helpful and detailed PR description.

Looks like a great addition.

tests/testthat/test-caption.R
kylebaron: Thanks; I've been including it but looks like testthat is using the file name if you don't include it which is what I wanted. But I should update to the new system at some point.

R/caption.R
kylebaron: Pushed some updates so that parser can handle newline and sensible embedded brackets.

R/table-object.R
kylebaron: I see. Thanks that's right. I think `short` used to be an argument to `st_caption()` but I moved it farther in. I'll get better documentation here.

R/caption.R
kylebaron: Good point. This was refactored out when I more carefully pull the short title out.

R/preview.R
kylebaron: The code isn't very clear; there's lots of different things using the same name. 

When you call `cap_short()` on 254, that will go into the table text, pull off the caption and try to pull off the short title. If there was no short title, that should come back null which is why the check is there. The stuff back on 217 - 221 is setting up captions and short names in case `st_caption()` wasn't used; that's the previous behavior that I tried to preserve.

R/preview.R
```diff
@@ -7,7 +7,10 @@ form_caption <- function(long = NULL, short = NULL, label = NULL) {
   if(is.character(label)) {
     label <- paste0(" \\", label)
   }
-  paste0("\\caption[",short,"]{",long,label,"}")
+  if(is.character(short)) {
+    short <- paste0("[", short, "]")
+  }
```
kylebaron, line 12: When `short` is `NULL`, the short name becomes `[]` and nothing shows up in toc. Refactoring this to only bracket when there is something to bracket.

R/preview.R
kyleam: > When you call cap_short() on 254 [...]

Doh, I missed that override.  Thanks.

R/caption.R
```diff
@@ -0,0 +1,153 @@
...
+  o <- order(positions)
+  bits <- bits[o]
+  positions <- positions[o]
+  done <- cumsum(bits)
+  done <- positions[done==0]
+  if(length(done)==0) {
+    if(mismatch) {
+      done <- positions[bits==-1][1]
+    } else {
+      return(list(main = x, short = ""))
```
kyleam, line 23: The tests don't hit this line, and I couldn't come up with a case where `done` would have no 0 elements but brackets were matched.  Am I missing one?

(If it should be inaccessible given upstream logic, signaling an error would make things clearer in my opinion, though I think it's okay as is.)

R/caption.R
```diff
@@ -0,0 +1,153 @@
...
+#' output file.
+#'
+#' @details
+#' Caption text (in `x`) can contain a short caption title enclosed in brackets
+#' (`[]`) as the opening characters (see examples). The short title is
+#' assumed to start after the opening `[` and end prior to `]` when all brackets
+#' are matched, allowing `[...]` in the short title as long as all brackets
+#' are matched. For other scenarios where a single `]` may need to appear
+#' in the short title, the user should use the `short` argument instead of
+#' the bracket notation instead.
```
kyleam, line 95: > the user should use the `short` argument instead of the bracket notation instead.

Repeated "instead".  Drop second one?

R/caption.R
kyleam: Nice, I just meant to note the limitations given that the `short` arg could be used, but your proposed approach looks good to me.

R/caption.R
kylebaron: Ugh

R/caption.R
kylebaron: I think this might be handled now by this bit upstream. I'll drop this whole block as unneeded now. 
```r
if(length(open)==0 || length(close)==0 || open[1] != 1) {
    return(list(main = x, short = ""))
  }
```

kylebaron: Ok ... revisiting this. Right ... so basically if `done` is never 0, then I'll just take the last close bracket and leave it at that. I think that's what you're addressing, right? The check of `mismatch` doesn't matter.

kyleam: > I think that's what you're addressing, right?

Yes, that takes care of what I was talking about (the unreachable `else` arm).

kyleam (approved):
Latest updates look good to me.

PR #303, opened 2022-12-09
silence tidyselect deprecation warnings

kyleam:
Running the test suite with the most recent tidyselect (1.2.0, released 2022/10) leads to a good number of warnings.  This series silences all of those.  I've confirmed that the tests pass with tidyselect 1.1.2 and 1.2.0.

More information: https://www.tidyverse.org/blog/2022/10/tidyselect-1-2-0/#using-data-inside-selections

PR #314, opened 2023-07-13
Clone an stobject

kylebaron:
- This is the pipeable table construction object
- It is an evironment, so special care when making copies


R/table-object.R
```diff
@@ -950,3 +950,33 @@ print.stobject <- function(x, ...) {
...
 #' @export
 names.stobject <- function(x) names(x$data)
+
+
+#' Clone an stobject
+#'
+#' @param x an stobject object.
+#'
+#' @return
+#' A deep copy of `x`.
```
kyleam, line 960: > A deep copy of `x`.

I think it'd be good to avoid the word "deep".  Perhaps I'm too influence by Python (which of course has very different semantics), but I'd expect a deep copy to be recursive.  I'm guessing there's no need for it to be recursive in this case, so my suggestion is just to adjust the docs.

<details>
<summary>example</summary>

For a deep copy, I'd expect the inner `aa` environment to also be copied.

``` r
st_clone <- function(x) {
  y <- list2env(as.list(x))
  attributes(y) <- attributes(x)
  y
}

a <- new.env(parent = emptyenv())
a$foo <- 1
aa <- new.env(parent = emptyenv())
aa$bar <- 2
a$aa <- aa

b <- st_clone(a)
b$foo <- b$foo * 10
b$new <- "abc"
b$aa$bar <- b$aa$bar * 10
b$aa$new <- "xyz"


b$foo
#> [1] 10
b$new
#> [1] "abc"
b$aa$bar
#> [1] 20
b$aa$new
#> [1] "xyz"

a$foo
#> [1] 1
a$new
#> NULL
a$aa$bar
#> [1] 20
a$aa$new
#> [1] "xyz"
```

</details>

R/table-object.R
```diff
@@ -950,3 +950,33 @@ print.stobject <- function(x, ...) {
...
+#' # Get back to a regular environment
+#' class(x) <- "environment"
+#' class(y) <- "environment"
+#'
+#' x
+#' y
+#'
+#' @export
+st_clone <- function(x) {
+  y <- list2env(as.list(x))
```
kyleam, line 979: For completeness, I think `as.list` should be called with `all.names = TRUE`.

Or perhaps use `rlang::env_clone` [*]?  That might take care of some remaining edge cases we're not thinking of and is faster (though I don't think it really matters in this context).

``` r
st_clone_1 <- function(x) {
  y <- list2env(as.list(x))
  attributes(y) <- attributes(x)
  y
}

st_clone_2 <- function(x) {
  y <- rlang::env_clone(x)
  attributes(y) <- attributes(x)
  y
}

a <- new.env(parent = emptyenv())
a$foo <- 1:10

rbenchmark::benchmark(
  st_clone_1(a),
  st_clone_2(a)
)
#>            test replications elapsed relative user.self sys.self user.child
#> 1 st_clone_1(a)          100   0.019     4.75     0.018    0.001          0
#> 2 st_clone_2(a)          100   0.004     1.00     0.004    0.000          0
#>   sys.child
#> 1         0
#> 2         0
```

[*] Note: `env_clone` doesn't do a recursive copy either.

kyleam:
Side remark: when running the tests, some deprecation warnings popped up.  The tidyselect ones are resolved by gh-303.

R/table-object.R
kylebaron: Fair enough. 

- Dropped "deep" from docs
- Using rlang clone function
- Will error if the argument isn't the object we were expecting (that would prevent the `example` you show from happening
- I believe (as the tests and examples show) once we "clone" this, the new object is subject to copy-on-modify behavior; I think some of the underlying objects in the clone and the original still point to the original address, but if you modify it will get copied
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(pmtables)
library(lobstr)
a <-
  stdata() %>%
  st_new() %>%
  st_center()


b <- st_clone(a)
c <- a

obj_addr(a)
#> [1] "0x143747dd8"
obj_addr(b)
#> [1] "0x1437be480"
obj_addr(c)
#> [1] "0x143747dd8"

obj_addr(a$align)
#> [1] "0x1164ee2e8"
obj_addr(b$align)
#> [1] "0x1164ee2e8"
b$align <- letters

str(a$align)
#> List of 5
#>  $ complete: NULL
#>  $ default : chr "c"
#>  $ update  : list()
#>  $ coltype : chr "p"
#>  $ outer   : chr "none"
#>  - attr(*, "class")= chr "aligncol"
obj_addr(a$align)
#> [1] "0x1164ee2e8"
obj_addr(b$align)
#> [1] "0x14406d430"
```

R/table-object.R
kylebaron: Thanks. I picked up `env_clone`

kyleam (approved):

PR #315, opened 2023-07-13
Adds update method for aligncol objects

kylebaron:
This lets you call `st_align()` or `st_center()` multiple times in a pipeline.

In the pipeline, we maintain an object with formatting information. Alignment is stored in `align` as an object with S3 class `aligncol`.  I created an update method for this object. The only hard thing to update is the `update` slot and there is already a `combine_list()` function that we use to update items in one list with items in another.

Checklist

- [x] Test


tests/testthat/test-align.R
```diff
@@ -14,3 +14,40 @@ test_that("align helpers [PMT-TEST-0001]", {
...
+  expect_identical(b$default, "l")
+
+  b <- update(a, .coltype = "m")
+  expect_is(b, "aligncol")
+  expect_identical(a$coltype, "p")
+  expect_identical(b$coltype, "m")
+
+  b <- update(a, .outer = "lr")
+  expect_is(b, "aligncol")
+  expect_identical(a$outer, "none")
```
kyleam, line 34: For completeness (and consistency with other assertions),  should there also be a `expect_identical(b$outer, "lr")`?

kyleam (approved):
> This lets you call st_align() or st_center() multiple times in a pipeline. [...]

Nice

PR #317, opened 2023-07-14
Keep row names in data.frame output from yaml_as_df

kylebaron:
# Summary 

`yaml_as_df()` reads in a yaml file and turns it into a data.frame. 

Yaml code might look like this:

```yaml
a: 
  name: THETA1
  parameter: CL
  trans: log
b: 
  name: THETA2
  parameter: V
  trans: log
```

We have to type the row names (`a` and `b`) but that information is lost when converted to data frame. 

This PR retains the row names in a `.row` column at the start of the data frame

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(pmtables)
options(pillar.width = Inf)

yaml <- '
THETA1:
  parameter: CL
  trans: log
THETA2:
  parameter: V
  trans: log
'
file <- file.path(tempdir(), "test.yaml")
writeLines(yaml, con = file)
yaml_as_df(file)
#> # A tibble: 2 × 3
#>   .row   parameter trans
#>   <chr>  <chr>     <chr>
#> 1 THETA1 CL        log  
#> 2 THETA2 V         log
```

Save as name instead

``` r
yaml_as_df(file, row_var = "name")
#> # A tibble: 2 × 3
#>   name   parameter trans
#>   <chr>  <chr>     <chr>
#> 1 THETA1 CL        log  
#> 2 THETA2 V         log
```

Opt out

``` r
yaml_as_df(file, row_var = NULL)
#> # A tibble: 2 × 2
#>   parameter trans
#>   <chr>     <chr>
#> 1 CL        log  
#> 2 V         log
```


R/yaml-as-df.R
```diff
@@ -53,10 +65,23 @@ yaml_as_df <- function(path, quiet = FALSE) { #nocov start
...
   if(exists("cols", meta)) {
     cols <- meta[["cols"]]
-    ans <- select(ans, cols, tidyselect::everything())
+    ans <- select(ans, all_of(cols), everything())
   }
+
+  if(is.character(row_var)) {
+    row_var0 <- row_var
+    i <- 1
+    while(row_var %in% names(ans)) {
```
kyleam, line 77: I like that this takes care to avoid a collision, but, as a user, I'd prefer an error here so that I don't get any surprises if my downstream code tries to use `row_name` while it really should be using, say, `row_name_1` due to a collision.

If the default of `row_var` was `NULL`, then that could be done without worrying about breaking existing code.  But given the default of `row_var = ".row"`, it'd be too risky.

Considering that ".row" is a nicer default (so that this behavior isn't disabled by default) and that collisions are likely to be rare, this logic to avoid the collision is probably the way to go.

R/yaml-as-df.R
```diff
@@ -53,10 +65,23 @@ yaml_as_df <- function(path, quiet = FALSE) { #nocov start
...
+      row_var <- paste0(row_var0, "_", i)
+      i <- i + 1
+    }
+    ans[, row_var] <- spec_names
+    ans <- select(ans, all_of(unique(c(row_var, names(ans)))))
```
kyleam, line 82: Okay as is, but some thoughts while reading along:

 * I wasn't able to come up with a case where the names and `row_var` wouldn't already be unique (given the upstream collision logic).

 * this `select` call is quite similar to the one on line 71, so as a reader it caused me to do a double take to see if there was any reason the approach on line 71 couldn't be used.  (I don't think there is.)

 * Two suggestions to simplify the "add row, rearrange row" into one step:

   ```
   ans <- tibble(!!row_var := spec_names, ans)
   ```

   or

   ```
   ans <- tibble::add_column(ans, !!row_var := spec_names, .before = 1)
   ```

tests/testthat/test-yaml_as_df.R
```diff
@@ -31,4 +31,33 @@ test_that("read prototyped table [PMT-TEST-0248]", {
...
+    .row: 1
+    .row_1: 11
+  aa:
+    b: 3
+    .row: 2
+    .row_1: 22
+  '
+  file <- file.path(tempdir(), "yaml-as-df-test-.row-rename")
+  writeLines(text = yaml, con = file)
+  df <- yaml_as_df(file)
```
kyleam, line 60: Using a temporary file is of course perfectly fine, but here's an approach with `textConnection()` that avoids touching the disk:

```
diff --git a/tests/testthat/test-yaml_as_df.R b/tests/testthat/test-yaml_as_df.R
index 2007667..4a81f5b 100644
--- a/tests/testthat/test-yaml_as_df.R
+++ b/tests/testthat/test-yaml_as_df.R
@@ -55,9 +55,7 @@ test_that(".row is renamed if it exists", {
     .row: 2
     .row_1: 22
   '
-  file <- file.path(tempdir(), "yaml-as-df-test-.row-rename")
-  writeLines(text = yaml, con = file)
-  df <- yaml_as_df(file)
+  df <- yaml_as_df(textConnection(yaml))
   expect_is(df, "data.frame")
   expect_true(names(df)[1] == ".row_2")
 })
```

kyleam (approved):
Seems like a good direction to go.  Left a few comments, but none of them are blockers.

tests/testthat/test-yaml_as_df.R
kylebaron: Thanks; I didn't know about this function. Super helpful.

R/yaml-as-df.R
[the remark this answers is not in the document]
kylebaron: I'm going to pass if it isn't character and error if it is character with length greater than 1

kylebaron: I wondered if this would be overkill.  

How about a message when we re-name the column with row names? I think it should be rare to see the message; but it would let the function to succeed and alert the user about it.

kyleam: Yeah, a message or warning seems reasonable given it's expected to be rare.

R/yaml-as-df.R
kylebaron: Thanks; went with the `tibble()` solution.

R/yaml-as-df.R
```diff
@@ -82,8 +84,10 @@ yaml_as_df <- function(path, quiet = FALSE, row_var = ".row") { #nocov start
       row_var <- paste0(row_var0, "_", i)
       i <- i + 1
     }
-    ans[, row_var] <- spec_names
-    ans <- select(ans, all_of(unique(c(row_var, names(ans)))))
+    if(!identical(row_var, row_var0)) {
+      message(glue("`{row_var0}` already exists; saving row names to `{row_var}`."))
+    }
+    ans <- as.data.frame(tibble(!!row_var := spec_names, ans))
```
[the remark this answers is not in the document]
kylebaron: Good point;  I usually try to return regular data frames, but wasn't in this case. I was mainly using `as_tibble()` to get the lists into data fames and should have coerced that to data frame instead in earlier iteration. But I think this has always returned tibble so I'll keep it that way. Thanks for picking that up.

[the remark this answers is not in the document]
kylebaron: I added an expectation that output is tibble for a reminder.

PR #326, opened 2024-02-15
Pull table notes from tex glossary files and yaml 

kylebaron:
# Summary

The PR adds functionality to create table notes from a tex glossary file. 

An entry in a glossary file looks like this

```tex
\newacronym{label}{abbreviation}{definition}
```
The basic idea is to read and parse the glossary file to create a string of the form `label: definition`. Likely multiple terms will be selected in which the string will take the form `label1: definition1; label2: definition2`. 

The PR also implements ability to read from glossary info in yaml-formatted file.

## Objects

- `glossary`  - a list of glossary entries; the names of the list are the glossary `labels`
- `glossary_entry` - a list containing the `abbreviation` and the `definition`

## Functions

- `read_glossary()` reads and parses the glossary file; returns a glossary object
- `glossary_notes()` takes in the glossary file name or a glossary list as well as labels to select and returns a character vector that can be added to a table via `st_notes()`
- `st_notes_glo()` takes a glossary list (from `read_glossary()`) and labels to select and adds the notes in a table pipeline
- `as_glossary()` coerce a list to a glossary object
- `update_abbrev()` - you can update the abbreviation for any entry (but can't change the label or the definition)

# Reprex

``` r
library(reprex)
library(pmtables)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
```

# Read in a glossary file

We can read from a `.tex` glossary file

``` r
glofile <- system.file("glo", "glossary.tex", package = "pmtables")

g <- read_glossary(glofile)
```

or a yaml file

``` r
gloyaml <- system.file("glo", "glossary.yaml", package = "pmtables")

y <- read_glossary(gloyaml)
```

The result is a glossary object

``` r
y
#> egfr : estimated glomerular filtration rate
#> bmi  : body mass index
#> wt   : weight
#> ht   : height
#> cmax : maximum concentration in the dosing inte...
#> cmin : minimum concentration in the dosing inte...
#> auc  : area under the concentration time curve
```

## yaml format

The yaml file needs to be written to be read by `yaml_as_df()`; the
outer level is the `labels` and inner are the `abbreviations` and
definitions

``` r
cat(readLines(gloyaml), sep = "\n")
#> egfr:
#>   def: estimated glomerular filtration rate
#>   abb: eGFR
#> bmi:
#>   def: body mass index
#>   abb: BMI
#> wt:
#>   def: weight
#>   abb: WT
#> ht:
#>   abb: HT
#>   def: height
#> cmax:
#>   def: maximum concentration in the dosing interval
#>   abb: Cmax
#> cmin:
#>   def: minimum concentration in the dosing interval
#>   abb: Cmin
#> auc:
#>   abb: AUC
#>   def: area under the concentration time curve
```

# Create a glossary object

``` r
x <- as_glossary(c = "cat", d = "dog", s = "snake")
x
#> c : cat
#> d : dog
#> s : snake
```

By default, the abbreviation is taken to be the label

# Update abbreviation

``` r
x <- update_abbrev(x, s = "SNAKE")
x$s
#> snake (SNAKE)
```

# Work with glossary object

A lot of this is driven by need to potentially combine glossary objects
or look into an object to see what is in there. I would have rather stuck
with a simpler data structure, but it needed to be more complex and I added
this functionality.

## Extract and print

``` r
g[1:10]
#> ADA    : anti-drug antibodies
#> AE     : adverse event
#> AIC    : Akaike information criterion
#> ALAG   : oral absorption lag time
#> ASCII  : American Standard Code for Information I...
#> AST    : aspartate transaminase
#> AUC    : area under the concentration-time curve
#> AUCss  : area under the concentration-time curve ...
#> AUCC   : cumulative area under the concentration-...
#> AUCC50 : area under the concentration-time curve ...

g$WT
#> subject weight (WT)
```

## Head

``` r
head(g)
#> ADA   : anti-drug antibodies
#> AE    : adverse event
#> AIC   : Akaike information criterion
#> ALAG  : oral absorption lag time
#> ASCII : American Standard Code for Information I...
#> AST   : aspartate transaminase
```

## Select

``` r
g2 <- select_glossary(g, AIC, AST, ADA)
```

## Combine

``` r
g3 <- c(g2, y)
```

### data frame

``` r
as.data.frame(g3)
#>    label                                   definition abbreviation
#> 1    AIC                 Akaike information criterion          AIC
#> 2    AST                       aspartate transaminase          AST
#> 3    ADA                         anti-drug antibodies          ADA
#> 4   egfr         estimated glomerular filtration rate         eGFR
#> 5    bmi                              body mass index          BMI
#> 6     wt                                       weight           WT
#> 7     ht                                       height           HT
#> 8   cmax maximum concentration in the dosing interval         Cmax
#> 9   cmin minimum concentration in the dosing interval         Cmin
#> 10   auc      area under the concentration time curve          AUC
```

### list

``` r
as.list(g3[1:2])
#> $AIC
#> $AIC$abbreviation
#> [1] "AIC"
#> 
#> $AIC$definition
#> [1] "Akaike information criterion"
#> 
#> 
#> $AST
#> $AST$abbreviation
#> [1] "AST"
#> 
#> $AST$definition
#> [1] "aspartate transaminase"
```

# Create notes

With a subset (expected most of the time)

``` r
glossary_notes(g3, AIC, wt, auc) 
#> [1] "AIC: Akaike information criterion; WT: weight; AUC: area under the concentration time curve"
```

With all entries

``` r
glossary_notes(g3)
#> [1] "AIC: Akaike information criterion; AST: aspartate transaminase; ADA: anti-drug antibodies; eGFR: estimated glomerular filtration rate; BMI: body mass index; WT: weight; HT: height; Cmax: maximum concentration in the dosing interval; Cmin: minimum concentration in the dosing interval; AUC: area under the concentration time curve"
```

In a pipeline

``` r
stdata() %>% 
  st_new() %>% 
  st_notes_glo(g3, AIC, wt, auc, width = 1) %>% 
  stable() %>% 
  st_as_image()
```

Alternatively

``` r
notes <- glossary_notes(g3, ht, wt, bmi)
stdata() %>% 
  st_new() %>% 
  st_notes(notes) %>% 
  st_panel("STUDY") %>% 
  stable() %>% 
  st_as_image()
```

Pass in names

``` r
labels <- c("AIC", "AST")
stdata() %>% 
  st_new() %>% 
  st_notes_glo(g3, labels = labels) %>% 
  stable() %>% 
  st_as_image()
```

<sup>Created on 2024-05-23 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


KatherineKayMRG:
@kylebaron I have a couple of questions about what this is doing (and if it's what we want). Above you say:

> An entry in a glossary file looks like this

```\newacronym{label}{abbreviation}{definition}```

> The basic idea is to read and parse the glossary file to create a string of the form `label: definition`. Likely multiple terms will be selected in which the string will take the form `label1: definition1`; `label2: definition2`.

So it's the information in the {label}{definition} that will go in the footer right? 

Can these functions handle cases where you might was the {abbreviation}{definition} combo? Maybe as an optional extra. For example, the CV% that you may have in the table usually uses CVP in the glossary:

```\newacronym{CVP}{CV\%}{percent coefficient of variation}```

I just had a similar case on a project where tables used FAPα but we couldn't use the greek letter in the glossary label, so the label was FAPa and the abbreviation included the greek letter

It would be nice to be able to specify whether label or abbreviation get used. 

Different question, will glossary entries like this (below) mess with your current code?

```\newacronym[sort=f]{F}{\ensuremath{F}}{absolute bioavailability}```

**Comment by kylebaron** [comment]
Hey @KatherineKayMRG  - 

Good points. I think we want to refer to to the `label`, but agree it would be better to put the `abbreviation` in there. I can make that happen. 


>Different question, will glossary entries like this (below) mess with your current code?
>
> \newacronym[sort=f]{F}{\ensuremath{F}}{absolute bioavailability}

I think the code handles this as it is.

**Comment by kylebaron** [comment]
I'm going to refactor this ... will be more complicated but I think there is a need.

File: R/glossary.R [hunk 1]
```diff
@@ -0,0 +1,390 @@
...
+read_tex_glossary <- function(file) {
+  txt <- readLines(file, warn = FALSE)
+  if(!length(txt)) abort("The glossary file was empty.")
+  parse_tex_glossary(txt)
+}
+
+parse_tex_glossary <- function(txt) {
+  txt <- trimws(txt)
+  txt <- txt[grepl("^\\\\newacronym", txt)]
+  m <- regexec("\\{(.+)\\}\\{(.+)\\}\\{(.+)\\}.*$", txt)
```
kyleam, line 112: I suspect `newacronym` items permit a fair amount of flexibility in spaces and newlines (e.g., between the brackets for different arguments or at the start of the line). The parser here doesn't allow that, but I think that's okay given that the expected forms are clearly documented.

Beyond whitespace, it looks like extra brackets could trip up the parser. Here's an example of brackets in a trailing comment causing issues:

```R
parse_tex_glossary("\\newacronym{a}{b}{c} % foo}")
#> $a
#> c} % foo (b)
```

Perhaps this parser should strip the trailing comments from the line before doing the match.

R/glossary.R
```diff
@@ -0,0 +1,390 @@
...
+#' @rdname glossary_notes
+#' @export
+glossary_notes.character <- function(x, ..., format = guess_glo_fmt(x)) {
+  x <- read_glossary(x, format = format)
+  glossary_notes(x, ...)
+}
+
+#' @rdname glossary_notes
+#' @export
+glossary_notes.list <- function(x, format = guess_glo_fmt(x), ...) {
```
kyleam, line 262: Is `format` a pasto from `glossary_notes.character`? Including `format` here will trigger an error:

```R
glossary_notes(list(ss = "steady state", ALB = "albumin", WT = "weight"))
#> Error in `guess_glo_fmt()` at pmtables/R/glossary.R:191:3:
#> ! Could not guess glossary format from file extension.
#> Run `rlang::last_trace()` to see where the error occurred.
#> Warning messages:
#> 1: In if (ext == "tex") return("tex") :
#>   the condition has length > 1 and only the first element will be used
#> 2: In if (ext %in% c("yml", "yaml")) return("yaml") :
#>   the condition has length > 1 and only the first element will be used
```

R/glossary.R
```diff
@@ -0,0 +1,390 @@
...
+  x <- as_glossary(x, format = format)
+  glossary_notes(x, ...)
+}
+
+#' @rdname glossary_notes
+#' @export
+glossary_notes.glossary <- function(x, ..., sep = ": ", collapse = "; ",
+                                    labels = NULL) {
+  labels <- cvec_cs(labels)
```
kyleam, line 271: What the motivation for including both `...` and `labels` as a way to select items? Is it just that it seems more intuitive than using `all_of`?

```R
labels <- c("ss", "ALB")
glossary_notes(g, all_of(labels))
#> "ss: steady state; ALB: albumin"
```

R/glossary.R
```diff
@@ -0,0 +1,390 @@
...
+  notes <- paste0(abb, sep, notes)
+  if(is.character(collapse)) {
+    notes <- paste0(notes, collapse = collapse)
+  }
+  notes
+}
+
+#' @export
+print.glossary_entry <- function(x, ...) {
+  cat(ans <- paste0(x$definition, " (", x$abbreviation, ")"))
```
kyleam, line 292: Unnecessary assignment?

kyleam, line 292: Nit-pick: it'd be good to include a newline at the end of this `cat` to play nicely with consoles other than RStudio that don't automatically tack on a missing newline. For example, with plain R from a terminal, the `>` gets stuck to the end of the output:

```
> as_glossary(list(ss = "steady state"))[[1]]
steady state (ss)>
```

R/glossary.R
```diff
@@ -0,0 +1,390 @@
...
+}
+
+#' @export
+print.glossary <- function(x, ...) {
+  label <- names(x)
+  def <- map_chr(x, "definition")
```
kyleam, line 298: This `print` method will fail if on an empty `glossary`:

```R
as_glossary(list(ss = "steady state"))[2]
#> Error in `map_chr()` at pmtables/R/glossary.R:298:3:
#> ℹ In index: 1.
#> Caused by error:
#> ! Result must be length 1, not 0.
#> Run `rlang::last_trace()` to see where the error occurred.
```

R/glossary.R
```diff
@@ -0,0 +1,390 @@
...
+  if(!identical(new_cols, names(y))) {
+    stop("`x` and `y` cannot share any names.")
+  }
+  ans <- structure(c(unclass(x), unclass(y)), class = cl)
+  more <- list(...)
+  if(length(more)) {
+    return(c(ans, more[[1]]))
+  }
+  ans
+}
```
kyleam, line 342: I'm confused by two related things here:

-   why only consider the first two items rather than concatenate an arbitrary number of items?

-   why take one the **first** item of `more`, and why process it differently than the first two items?

I'm thinking instead the definition should look something like this:

```R
#' @export
c.glossary <- function(..., recursive = FALSE) {
  args <- list(...)
  if (!length(args)) {
    return(NULL)
  }
  if (!all(map_lgl(args, function(x) inherits(x, "glossary")))) {
    stop("All arguments must be glossary objects.")
  }

  cl <- class(args[[1]])
  n <- sum(map_int(args, length))
  nms <- unique(unlist(lapply(args, names)))
  if (n > length(nms)) {
    stop("Arguments cannot share any names.")
  }

  res <- do.call(c, lapply(args, unclass))
  return(structure(res, class = cl))
}
```

R/glossary.R
```diff
@@ -0,0 +1,390 @@
...
+  unclass(x)[[name]]
+}
+
+#' @export
+`[.glossary` <- function(x, i, j, drop = FALSE) {
+  cl <- class(x)
+  x <- unclass(x)
+  x <- x[i]
+  structure(x, class = cl)
+}
```
kyleam, line 358: Could this be simplified to this?

```R
#' @export
`[.glossary` <- function(x, ...) {
  structure(NextMethod(), class = class(x))
}
```

R/glossary.R
```diff
@@ -0,0 +1,390 @@
...
+  ans
+}
+
+#' @export
+`$.glossary` <- function(x, name, ...) {
+  if(!exists(name, x)) {
+    return(NULL)
+  }
+  unclass(x)[[name]]
+}
```
kyleam, line 350: Just to check my understanding: you're defining a custom method (rather than letting the parent method handle it) to avoid matching on partial input, right?

I think it'd be good to add a test for the "error on partial input" behavior.

tests/testthat/test-glossary.R
```diff
@@ -0,0 +1,259 @@
...
+  expect_identical(names(x), "a")
+  expect_equivalent(
+    x$a, list(abbreviation = "b_\\mathrm{d}", definition = "\\mathit{c}")
+  )
+
+  txt <- "%\\newacronym{a}{b}{c}"
+  expect_error(pmtables:::parse_tex_glossary(txt), "No acronym entries")
+})
+
+test_that("corece list to glossary object", {
```
kyleam, line 138: typo: corece

kyleam (changes requested):

R/glossary.R
kylebaron: Thanks, Kyle. 

1. spaces at the start of a line are handled by `trimws()`
2. Modified the regex to allow for spaces between brackets
3. Trimmed parsed result to take care of spaces inside brackets at the start or end
4. Stripped trailing comments; this starts at the first unescaped `%`

Tests added for this but please let me know if something is still off.

R/glossary.R
kylebaron: Good idea; I didn't think about this.

R/glossary.R
kylebaron: Thanks; I think this is a mash up of some different stuff I tried.

kylebaron: Never thought of this before.

R/glossary.R
kylebaron:
```r
> l <- as_glossary(a = "b")[0]
> l
No glossary entries found.
```

R/glossary.R
kylebaron: It wasn't partial matching; I thought it there was some issue extracting via the `list` method. But I removed the `glossary` method and it seems to be working fine now. Will go with it.

kylebaron: I think it was the `[[` method that maybe dropped the class attribute of the thing you extracted. But not totally sure right now. I'd be happy to remove both single and double bracket methods if they are not needed.

R/glossary.R
kylebaron: Thanks; much simpler and I think it retains the class attribute of the pre-subset object which is what set a lot of this off.

R/glossary.R
kylebaron: Thanks. This works great and more straightforward than what I had, which was based on some misguided ideas on S3 dispatch (just didn't think it through).

R/glossary.R
kylebaron: I can't explain it either. Fixed this signature and the function; added tests for making glossary notes from a list. The original was hubris on my part. But I was on a roll ...

R/glossary.R
kyleam: > spaces at the start of a line are handled by `trimws()`

Bah, thanks, I had that straight at one point, and then got it wrong with a last minute edit to my comment.

kylebaron:
@kyleam  - I _think_ I addressed everything here, adding tests where I missed badly on some of the implementation. But let me know if I didn't get one of the comments right or if I just overlooked anything.

KatherineKayMRG (approved):
From a user perspective, this looks good to me and I'm looking forward to using it. No more questions from me. I'll leave the code review to Kyle M.

kylebaron:
Thanks, @KatherineKayMRG. It sounds like @timwaterhouse  is going to give this a spin on upcoming project and we can tweak and adjust some things from there.

R/glossary.R
```diff
@@ -0,0 +1,398 @@
...
+
+#' @export
+print.glossary <- function(x, ...) {
+  if(!length(x)) {
+    cat("No glossary entries found.")
+    return(invisible(NULL))
+  }
+  label <- names(x)
+  def <- map_chr(x, "definition")
```
kyleam, line 311: The case I mentioned [here](https://github.com/metrumresearchgroup/pmtables/pull/326#discussion_r1621588141) will still fail:

```R
as_glossary(ss = "steady state")[2]
#> Error in `map_chr()` at pmtables/R/glossary.R:311:3:
#> ℹ In index: 1.
#> Caused by error:
#> ! Result must be length 1, not 0.
#> Run `rlang::last_trace()` to see where the error occurred
```

The length guard won't catch that because of this named-list behavior:

```R
x <- list(foo = "foo")
x[2]
#> $<NA>
#> NULL

length(x[2])
#> [1] 1
```

How about this guard instead?

```diff
diff --git a/R/glossary.R b/R/glossary.R
index 76fc618..3016dc9 100644
--- a/R/glossary.R
+++ b/R/glossary.R
@@ -303,8 +303,8 @@ print.glossary_entry <- function(x, ...) {

 #' @export
 print.glossary <- function(x, ...) {
-  if(!length(x)) {
-    cat("No glossary entries found.")
+  if(is.null(unlist(x))) {
+    cat("No glossary entries found.\n")
     return(invisible(NULL))
   }
   label <- names(x)
```

kyleam (approved):
Thanks for the updates.  I have one follow-up comment about a case where `print.glossary` still errors, but otherwise this looks ready to go.

(I see there are merge conflicts, but those are just in the generated coverage reports.)

PR #328, opened 2024-02-29
Set option for maxex argument to sig

kylebaron:
sig renders numbers as character strings with a certain number of significant digits. Big numbers can get represented in scientific notation and we can control that through the `maxex` argument. Frequently (most of the time) we don't want scientific notation and if creating this function today, I'd probably default `maxex` to some large number. 

This PR lets you control `maxex` through an option allowing users to turn off scientific notation globally in a project.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -46,7 +46,7 @@ cvec_cs <- function(x) {
 #' @md
 #' @rdname sig
 #' @export
-sig <- function(x, digits = 3, maxex = NULL, ...) {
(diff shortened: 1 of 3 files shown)


KatherineKayMRG (approved):
Thanks Kyle - very helpful addition!

kyleam (approved):
Exposing as option makes sense to me.

This PR branched before the GitHub Actions switch.  Here's the passing [Drone build](https://github-drone.metrumrg.com/metrumresearchgroup/pmtables/1650).  I'll check back once this is merged and make sure the Actions for `main` pass.

PR #333, opened 2024-05-20
Feature - save table as image

kylebaron:
# Summary

This PR provides a single wrapper to save a table out to `png` or `pdf` format. It is set up to work like `stable_save()`, with some modification.


R/preview-standalone.R
```diff
@@ -436,3 +436,83 @@ st_image_show <- function(path,
...
+#' @param stem used to form the final output file name; this argument will
+#' override what is provided in `file`.
+#' @param format the output file format.
+#' @param dir the final output directory.
+#' @param build_dir the directory where the image will be built.
+#' @param ... passed to [st_aspdf()] or [st_aspng()], depending on the value of
+#' `format`.
+#'
+#' @details
+#' The table will be built / rendered in `tempdir()`.
```
kyleam, line 454: This seems confusing given that it depends on the `build_dir` argument. I think it'd be fine to drop this line.

R/preview-standalone.R
```diff
@@ -436,3 +436,83 @@ st_image_show <- function(path,
...
+#'
+#' @examples
+#' tab <- stable(stdata(), output_file = "foo.tex")
+#'
+#' \dontrun{
+#'   ans <- stable_save_image(tab format = "pdf")
```
kyleam, line 460: Missing comma after `tab`.

R/preview-standalone.R
```diff
@@ -436,3 +436,83 @@ st_image_show <- function(path,
...
+#'
+#' @seealso [st_as_image()]
+#'
+#' @export
+stable_save_image <- function(x,
+                              file = attr(x, "stable_file"),
+                              stem = NULL,
+                              format = c("png", "pdf"),
+                              dir = getOption("pmtables.dir"),
+                              build_dir = tempdir(), ...) {
```
kyleam, line 473: Plain `tempdir()` calls that give the top-level per-session temporary directory, instead of a within-session unique directory like you'd get with `tempfile()~/~withr::local_tempdir()`, make me a bit nervous, but it looks like this matches the behavior of `st_aspdf()` and `st_aspng()`, so it seems it's not a problem in this context and should be fine here.

R/preview-standalone.R
```diff
@@ -436,3 +436,83 @@ st_image_show <- function(path,
...
+      "maybe this object was corrupted or it wasn't generated from 'stable()'",
+      call.=FALSE
+    )
+  }
+  format <- match.arg(format)
+  if(is.character(file)) {
+    ext <- file_ext(file)
+    if(ext=="tex") {
+      stem <- file_path_sans_ext(file)
+      file <- paste0(stem, format)
```
kyleam, line 487: -   For "foo.tex" and `format = "pdf"`, this would incorrectly produce "foopdf".

-   That ends up not mattering because, in this `ext=="tex"` case, `stem` is a character, so the `file <- paste0(stem, ".", format)` on line 491 overrides this value.

So it seems line 487 can be dropped.

R/preview-standalone.R
```diff
@@ -436,3 +436,83 @@ st_image_show <- function(path,
...
+  } else {
+    ans <- st_aspng(
+      x,
+      stem = file_path_sans_ext(file),
+      dir = build_dir,
+      ...
+    )
+  }
+  if(!is.null(dir)) {
+    file <- file.path(dir, file)
```
kyleam, line 512: Based on the logic on this line and light testing, I think the expectation is that `file` is just a base name and never contains any directory components. Should this function guard against that and give a more informative error?

```
stable_save_image(stable(stdata()), file = "/tmp/foo.pdf")
#> Warning messages:
#> 1: In system2(command = command, args = args, stdout = TRUE, stderr = TRUE) :
#>   running command ''latex' -halt-on-error -jobname=/tmp/foo standalone-preview.tex 2>&1' had status 1
#> 2: In st_to_standalone(x, stem, dir, command = "latex", font = font,  :
#>   the standalone preview build didn't work.
#> 3: In file.create(to[okay]) :
#>   cannot create file '/tmp/Rtmpvtnp6Q//tmp/foo.png', reason 'No such file or directory'
```

kyleam (changes requested):
Looks good to me overall.  Two main comments:

* syntax error in doc examples (line 457)
* dead code that sets `file` to incorrect name that later gets reassigned to the intended (line 487)

---

> Should we warn for long table?

Is the idea that for `longtable`, we're breaking across multiple pages, so it's not useful to cram all that into an image?  Or are there other issues with the output?

> Does the output file name approach make sense? It's a mix of what we have in `stable_save()` and `mrggsave()`; I do think we'd like flexibility somewhere along these lines

My opinion shouldn't count for much on this, as I've never worked with either of those, but the approach makes sense to me.

R/preview-standalone.R
kylebaron: Thanks; this was left over from previous iteration where we didn't give `build_dir` option.

R/preview-standalone.R
kylebaron: Thanks; I noticed something wasn't behaving as I had expected when showing @KatherineKayMRG the code yesterday. I think this is the cause. I'll refactor this and confirm the behavior.

R/preview-standalone.R
```diff
@@ -436,3 +436,108 @@ st_image_show <- function(path,
...
+#' @details
+#'
+#' There is considerable flexibility for specifying the output file name. This
+#' flexibility is provided to accommodate several potential different
+#' anticipated workflows.
+#'
+#' - `file` can be obtained from the `stable_file` attribute on `x`; this
+#'   attribute is set when the user specifies the output file name via
+#'   `st_files()`; if a `.tex` extension is found on `file`, it will be
+#'   changed to the value found in `format`; in this case, in this case, `file`
```
kyleam, line 464: repeated "in this case"

R/preview-standalone.R
```diff
@@ -436,3 +436,108 @@ st_image_show <- function(path,
...
+#' Render and save a table in png or pdf format
+#'
+#' @param x an stable object.
+#' @param file the final output file name; if `file` has `.tex` extension, the
+#' extension will be replaced with the value in `format`; if provided as a
+#' complete path, directory information will be used as `dir` (see below).
+#' @param stem used to form the final output file name; this argument will
+#' override what is provided in `file`.
+#' @param format the output file format.
+#' @param dir the final output directory; will be overwritten if `file` is
```
kyleam, line 449: Maybe "overridden" instead of "overwritten"? I'm worried that someone reading this quickly will interpret it as the directory itself being overwritten rather than `dir` being set to the value of `file`s directory.

kyleam (approved):
LGTM aside from doc typo.

R/preview-standalone.R
kylebaron: Thanks; good suggestion ... I didn't consider this.

R/preview-standalone.R
kylebaron: Thanks.

PR #341, opened 2024-06-20
Optionally print project path in table notes annotation

kylebaron:
# Summary

- Retain the current behavior of writing the name of the output file into the table annotation
- As noted in the attached vignette, the big change here is ability to add both the output file name and output path (directory) into the table annotation; this means we need to lock in the final location for the table as we build the table; the vignette shows this.
- `tab_notes()` gains the following arguments
  - `output_dir` where (what directory) to save the file; respects `pmtables.dir`
  - `path.type` how to render `output_dir` in the annotation; 
  - See implementation details
- New function: `format_table_path()`; does equivalent of what we already do in `mrggsave`

# Current default

Currently, we can take in the output file name and that gets 
included in a table note. However, the decision about _where_ 
to save that file is deferred until we actually call 
`stable_save()`. 

The changes to pmtables include accepting the output file path 
when we create the table. Now, if the user passes the output path 
and it is included in the table annotation, we have to respect that
information when saving; it can't be changed after the annotation 
is created. 

## Implementation detail

Currently, it is the `tab_notes()` function that handles all the 
arguments for R source file, output file, output directory etc. But 
this is a but of a misnomer now, because the function doing more 
than just writing a note and saving the file name for later; real
decisions about output file locations are being made. 

I have added a non-exported function `tab_files()` that handles 
all the file / path manipulation and arguments that get passed
to `tab_notes()` eventually go to `tab_files()`. In the future, 
I will open up `tab_files()` since that more accurately reflects
where the file / path proecessing is taking place. 

[source-path.pdf](https://github.com/user-attachments/files/15917796/source-path.pdf)


R/utils.R
```diff
@@ -176,3 +176,67 @@ ensure_parens <- function(x) {
...
+#' to an RStudio project file; use `"none"` to format without any directory
+#' information (just the output file name); use `"raw"` to print the complete
+#' path to the file using `dir` as-is.
+#'
+#' @details
+#' Note that the default value for `path.type` is `"proj"`. This requires
+#' that an RStudio project file is able to be found using
+#' [rprojroot::find_root()] with the [rprojroot::is_rstudio_project] criterion.
+#'
+#' Once mrggsave finds a root for a given working directory, it caches the value
```
kyleam, line 194: `s/mrggsave/pmtables/`

R/utils.R
```diff
@@ -176,3 +176,67 @@ ensure_parens <- function(x) {
...
+#' for the remainder of the R session. An error will be generated if an image is
+#' attempted to be saved using `path.type="proj"` but an RStudio project file
+#' was not able to be located.
+#'
+#' @return
+#' A string with the formatted image file path.
```
kyleam, line 200: Is "image" pasto from mrggsave?

R/table-notes.R
```diff
@@ -52,12 +60,55 @@ tab_notes <- function(notes, escape_fun = tab_escape,
...
+  out <- list(
     m_notes = m_notes, t_notes = t_notes, config = note_config,
-    notes = notes, r_file = r_file, output_file = output_file
+    notes = notes
   )
+  c(out, file_info)
 }
 
+# Defaults for `output_file` and `output_dir` are provided by
+# `table_notes()`
```
kyleam, line 71: `s/table_notes/tab_notes/`

kyleam:
I'll interactively test this a bit, but this is looking good to me aside from my minor comments above.

KatherineKayMRG:
From a user perspective, this worked well for me on the AZ project and I like the opt in approach.

KatherineKayMRG (approved):
From a user perspective, this worked well for me on the project and I like the opt in approach. I'll leave the more thorough code review to @kyleam

kyleam (approved):
LGTM aside from typos mentioned above

kylebaron:
Thanks, @kyleam ; working on the typos now.

PR #355, opened 2025-06-04
Caption as conditional macro

kylebaron:


R/table-long.R
```diff
@@ -3,6 +3,20 @@ longtable_head <- function(multicol) {
   c("\\endhead", "\\hline", multicol, "\\endfoot", "\\hline", "\\endlastfoot")
 }
 
+conditional_macro <- function(macro_name) {
+  macro <- paste0("\\", macro_name)
+  ans <- c(
+    "",
+    paste0("\\ifdefined", macro),
```
luislomeli-metrumrg, line 10: Does this allow that `macro_name` could be defined somewhere during the table specification (how?) and use it here?

kylebaron: @luislomeli-metrumrg  - 

Yes; `macro_name` here is whatever you pass into `stable_long()` as `lt_cap_macro`.  So you give it the name of the macro, and pmtables will resolve that macro in the caption text that must be included _in_ the longtable environment. The code that I've included here just wraps the sourcing of that macro in some conditional text such that if the macro is not defined at the time the table is rendered, it will give this boilerplate which tells you the macro will be used as soon as it gets defined.

kylebaron:
@kyleam  - 

If I'm reading this right, we can't (fully) test agains R 4.1 now?
https://github.com/metrumresearchgroup/pmtables/actions/runs/15454436675/job/43582234127?pr=355#step:4:214

R/table-long.R
luislomeli-metrumrg: Great, thanks for the detailed answer

luislomeli-metrumrg (approved):
LGTM, pending only the R 4.1 CI checks

kyleam:
@kylebaron I've pushed an update to pin both magick and scales in the R 4.0 jobs to the last versions that didn't require R 4.1.

kyleam:
I said:
> I've pushed an update to pin both magick and scales in the R 4.0 job to the last versions that didn't require R 4.1.

@kylebaron Since my last push, a few more packages have become incompatible with R 4.1 [*].  I'll push an update to get the R 4.0 job passing.  Assuming that's green, it looks like this is ready to merge?

[*] The R 4.0 list of pins is getting a bit long.  The next time we need to update it, it's worth considering whether pmtables needs to maintain compatibility with R <= 4.0 at this point.  If it does, it'd probably make sense to freeze the entire set of packages for the R 4.0 job (by switching to a https://packagemanager.posit.co/cran/__linux__/jammy/YYYY-MM-DD source) so that we won't need to keep updating the list of pins.

kylebaron:
Thanks @kyleam. Will merge this.

PR #373, opened 2026-01-21
Address missing values in the by column under pt_data_inventory

### LINKED ISSUE #372: pt_data_inventory labels NA as "Group Total"
If the "by" variable includes NA, pt_data_inventory lists them as separate group, but labels then as "Group Total" rather than"missing" or NA.

Image

kylebaron:
See #372 

The fix will be to call `replace_na()` on the `by` column prior to binding on the "total" row. 

TODO: 

- [x] Add test

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: pmtables
 Type: Package
 Title: Tables for Pharmacometrics
-Version: 0.10.0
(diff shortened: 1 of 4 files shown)


kyleam (approved):
LGTM

---

interactive confirmation of test case

**before**

20260121T144804

**after**

20260121T144635

PR #378, opened 2026-03-05
Refactor sig - big.mark and default maxex

kylebaron:
- `maxex` defaults to `Inf`
- Optionally use `big.mark` in `formatC()` only when >= 5 digits
  - Make two passes to properly handle big.mark in a vector

NOTE: I started thinking a bit more about how `formatC()` decides when to put numbers into scientific notation. @kyleam could you look at this and see what you think? So far, it works with the tests I have.  The goal of the update is to get rid of regex to figure out how to render the numbers. 

EDIT: I think I have this nailed down better now; legacy sig added; filled in some tests.

``` r
library(pmtables)

sigg <- function(..., maxex = Inf) pmtables:::sig_legacy(..., maxex = maxex)

sig(1.98765)
#> [1] "1.99"
sig(12.98765)
#> [1] "13.0"
sig(123.98765)
#> [1] "124"
sig(1234.98765)
#> [1] "1230"
sig(9994.9999)
#> [1] "9990"
sig(9995.9999)
#> [1] "10000"
sig(12345.98765)
#> [1] "12300"
sig(123456.98765)
#> [1] "123000"
sig(1234567.98765)
#> [1] "1230000"

sig(123, digits = 3)
#> [1] "123"
sig(1234, digits = 3, maxex = Inf)
#> [1] "1230"
sig(1234, digits = 3, maxex = 2)
#> [1] "1230"
sig(1234, digits = 3, maxex = 3)
#> [1] "1.23e+03"
sig(1234, digits = 3, maxex = 4)
#> [1] "1230"


x <- c(0.1, 1, 10, 100, 1000, 10000)
sig(x, maxex = 3)
#> [1] "0.100"    "1.00"     "10.0"     "100"      "1.00e+03" "1.00e+04"
sig(x, maxex = 3, digits = 2)
#> [1] "0.10"    "1.0"     "10"      "100"     "1.0e+03" "1.0e+04"
sigg(x, maxex = 3, digits = 2)
#> [1] "0.10"    "1.0"     "10"      "100"     "1.0e+03" "1.0e+04"


sig(0.01234567)
#> [1] "0.0123"
sig(0.001234567)
#> [1] "0.00123"
sig(0.0001234567)
#> [1] "0.000123"
sig( 0.00001234567)
#> [1] "0.0000123"
sigg(0.00001234567)
#> [1] "0.0000123"
sigg(0.00000000012321)
#> [1] "0.000000000123"
sig(0.00000000012321)
#> [1] "0.000000000123"

sig(0.000001234567)
#> [1] "0.00000123"
sig(0.00001234567, maxex = 5)
#> [1] "1.23e-05"
sig(0.00001234567, maxex = 6)
#> [1] "0.0000123"


sig(1234567.98765, maxex = 4, big.mark = ",")
#> [1] "1.23e+06"
sig(1234567.98765, maxex = 5, big.mark = ",")
#> [1] "1.23e+06"
sig(1234567.98765, maxex = 6, big.mark = ",")
#> [1] "1.23e+06"
sig(1234567.98765, maxex = 7, big.mark = ",")
#> [1] "1,230,000"
x <- c(1,10,100,1000,10000)
sig(x, big.mark = ",")
#> [1] "1.00"   "10.0"   "100"    "1000"   "10,000"

sig(0.1234567, maxex = 1)
#> [1] "0.123"
sig(0.1234567, maxex = 7)
#> [1] "0.123"
sig(0.000001234567, maxex = 6)
#> [1] "1.23e-06"
sig(0.000001234567, maxex = 7)
#> [1] "0.00000123"
sig(0.1234567, maxex = 70)
#> [1] "0.123"
```

<sup>Created on 2026-04-13 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -39,6 +39,7 @@ cvec_cs <- function(x) {
 #' sig(1.123455, digits = 5)
 #' sig(1123, maxex = 3)
 #' sig(1123, maxex = 4)
+#' sig(1123, big.mark = ",")
 #'
 #' sig(1L)
 #'
@@ -60,10 +61,11 @@ sig <- function(x, digits = 3,
 
   x <- as.numeric(x)
 
-  sigx <- signif(x, digits = 3)
+  sigx <- signif(x, digits = digits)
 
   bigm <- ifelse(is.character(big.mark) && sigx >= 10000, big.mark, "")
 
+  # This adds scientific notation; but also gets zeros right
   ans <- formatC(
     sigx,
     digits = digits,
@@ -72,15 +74,19 @@ sig <- function(x, digits = 3,
     big.mark = bigm
   )
 
+  # This looks for numbers in scientific notation
   if(is.numeric(maxex)) {
     if(digits != maxex) {
       ex <- "([-]*[0-9]\\.[0-9]+)e([+-][0-9]{2})"
-      subit <- grepl(ex, ans, perl = TRUE)
-      b <- as.numeric(gsub(ex, "\\2", ans))
+      subit <- vector(mode = "logical", length = length(ans)) 
+      e <- grepl("e", ans, fixed = TRUE)
+      subit[e] <- grepl(ex, ans[e], perl = TRUE)
+      b <- rep(Inf, length(ans))
+      b[e] <- as.numeric(gsub(ex, "\\2", ans[e]))
       subit <- subit & abs(b) < maxex
-      if(subit) {
-        ans <- formatC(
-          sigx,
+      if(any(subit)) {
+        ans[subit] <- formatC(
+          sigx[subit],
           digits = digits,
           format = "fg",
           flag = "#",
--- a/R/utils.R
+++ b/R/utils.R
@@ -66,16 +66,25 @@ sig <- function(x, digits = 3,
 
   sigx <- signif(x, digits = digits)
 
-  bigm <- ifelse(is.character(big.mark) && sigx >= 10000, big.mark, "")
+  big <- sigx >= 10000
 
   # This adds scientific notation; but also gets zeros right
   ans <- formatC(
     sigx,
     digits = digits,
     format = "g",
-    flag = "#",
-    big.mark = bigm
+    flag = "#"
   )
+
+  if(is.character(big.mark)) {
+    ans[big] <- formatC(
+      sigx[big],
+      digits = digits,
+      format = "g",
+      flag = "#", 
+      big.mark = big.mark
+    )
+  }
 
   # This looks for numbers in scientific notation
   if(is.numeric(maxex)) {
@@ -92,9 +101,17 @@ sig <- function(x, digits = 3,
           sigx[subit],
           digits = digits,
           format = "fg",
-          flag = "#",
-          big.mark = bigm
+          flag = "#"
         )
+        if(is.character(big.mark)) {
+          ans[subit & big] <- formatC(
+            sigx[subit & big],
+            digits = digits,
+            format = "fg",
+            flag = "#", 
+            big.mark = big.mark
+          )
+        }
       }
     }
   }
--- a/R/utils.R
+++ b/R/utils.R
@@ -26,6 +26,8 @@ cvec_cs <- function(x) {
 #' @param digits `numeric`; number of significant digits.
 #' @param maxex `numeric`; maximum number of significant
 #' digits before moving to scientific notation.
+#' @param big.mark if `character`, passed to [formatC()], but only for numbers
+#' greater than or equal to 10000 (five digits before decimal).
 #' @param ... other arguments that are not used.
 #'
 #' @return
@@ -46,7 +48,9 @@ cvec_cs <- function(x) {
 #' @md
 #' @rdname sig
 #' @export
-sig <- function(x, digits = 3, maxex = getOption("pmtables.maxex", Inf), ...) {
+sig <- function(x, digits = 3,
+                maxex = getOption("pmtables.maxex", Inf),
+                big.mark = getOption("pmtables.big.mark", NULL), ...) {
 
   if(identical(class(x), "integer")) {
     return(as.character(x))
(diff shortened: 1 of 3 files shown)


kylebaron:
Save: 
```r
      #ex <- "([-]*[0-9]\\.[0-9]+)e([+-][0-9]{2})"
      #subit <- vector(mode = "logical", length = length(ans))
      #e <- grepl("e", ans, fixed = TRUE)
      #subit[e] <- grepl(ex, ans[e], perl = TRUE)
      #b <- rep(Inf, length(ans))
      #b[e] <- as.numeric(gsub(ex, "\\2", ans[e]))
      #subit <- subit & abs(b) < maxex

```

R/utils.R
```diff
@@ -46,7 +54,73 @@ cvec_cs <- function(x) {
...
+    )
+  }
+
+  # This looks for numbers in scientific notation
+  if(is.numeric(maxex)) {
+    if(digits != maxex) {
+      ex <- log10(abs(sigx))
+      ex <- ifelse(ex < 0, ceiling(abs(ex)), floor(abs(ex)))
+      # We revert these numbers back to standard notation
+      subit <- x != 0  & (ex < maxex | (ex >= maxex & ex <= digits))
```
kylebaron: I _believe_ this will identify numbers that are rendered with an exponent and that need to be converted back to standard notation while also considering `maxex`.

kylebaron: Tweaked the subit logic: 

```r
subit <- x != 0  & (ex < maxex | (ex >= maxex & ex < digits))
```

R/utils.R
```diff
@@ -46,7 +54,74 @@ cvec_cs <- function(x) {
...
+
+  if(dobig) {
+    ans[big] <- formatC(
+      sigx[big],
+      digits = digits,
+      format = "g",
+      flag = "#",
+      big.mark = big.mark
+    )
+  }
```
kyleam, line 90: This block isn't covered in a meaningful way by the current tests (i.e., if it's deleted, all tests still pass) because a non-default `big.mark` isn't specified when `maxex` is `NULL` or equal to `digits`.

How about adding two tests like below?

```R
test_that("sig big.mark works with maxex=NULL", {
  x <- c(1000, 10000, 100000)
  a <- sig(x, digits = 8, maxex = NULL, big.mark = "_")
  expect_equal(a, c("1000.0000", "10_000.000", "100_000.00"))
})

test_that("sig big.mark works with maxex=digits", {
  x <- c(1000, 10000, 100000)
  a <- sig(x, digits = 8, maxex = 8, big.mark = " ")
  expect_equal(a, c("1000.0000", "10 000.000", "100 000.00"))
})
```

R/utils.R
```diff
@@ -46,7 +54,74 @@ cvec_cs <- function(x) {
...
+
+  # This looks for numbers in scientific notation
+  if(is.numeric(maxex)) {
+    if(digits != maxex) {
+      ex <- log10(abs(sigx))
+      ex2 <- ifelse(ex < 0, ceiling(abs(ex)), floor(abs(ex)))
+      # We revert these numbers back to standard notation
+      subit <- x != 0  & (ex2 < maxex | (ex2 >= maxex & ex2 < digits))
```
kyleam: The tests still pass if this is reduced to `subit <- ex2 < maxex`.  It'd be good to have cases to cover the other conditions.

R/utils.R
kylebaron: Thanks; tests added and passing.

R/utils.R
kylebaron: Thanks; test added that exercises the "or" condition on that `subit` statement.

kyleam: Thanks.  The new test does trigger the execution of that condition but not in a way that shows when the condition is necessary.  It passes if I remove the condition (i.e. if `subit` comes up as `FALSE`).

```
$ git rev-parse HEAD
c948fac57be5e8f96c03b9eca1c185d1b8b8bcc4

$ git diff
diff --git a/R/utils.R b/R/utils.R
index 68fd802..8a9d954 100644
--- a/R/utils.R
+++ b/R/utils.R
@@ -98,7 +98,7 @@ sig <- function(x, digits = 3,
       ex <- ifelse(ex < 0, ceiling(abs(ex)), floor(abs(ex)))

       # We revert these numbers back to standard notation
-      subit <- ex < maxex | (ex >= maxex & ex < digits)
+      subit <- ex < maxex

       if(any(subit)) {
         ans[subit] <- formatC(

$ Rscript -e 'devtools::test(filter="utils")'
ℹ Testing pmtables
✔ | F W  S  OK | Context
✔ |         70 | test-utils-table.R
✔ |         68 | test-utils

══ Results ═════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Duration: 1.0 s

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 138 ]
```

Is there a case when the condition is necessary?

kylebaron: Updated to remove the "or" `subit` condition. Passing check still.

kyleam (approved):
I don't spot any issues with the new logic (and of course the tests against the old implementation are reassuring).

I suspect that you, @KatherineKayMRG, and others have discussed the `maxex` default change.  For posterity, I think it'd be good for this PR to include the rationale for that change in behavior.

kylebaron:
Thanks, @kyleam. Yeah, this could have been much simpler (just change the default) but couldn't resist thinking about the implementation more and trying to improve. This approach was originally a hacky way to fix some outputs on a project and we never really thought about it more, how `formatC()` was working etc. I feel like a lot comes down to how that `subit` evaluation is set up and that turns out to be fairly simple at the end of the day. The tests have been greatly expanded here and a lot of different edge cases were explored and, yeah, to the extent that the legacy implementation did what we wanted it should be ok. 

We did discuss the default change on `slack` and there seemed to be wide support for the new default. We also discussed integration with `pmparams`, which imports this function. So we've thought through some of larger questions coming up to this. Thanks again for looking so closely at the code / logic / tests.

kylebaron:
@KatherineKayMRG what say you? is this good to go?

KatherineKayMRG:
LGTM

KatherineKayMRG (approved):

kylebaron:
Thanks @kyleam and @KatherineKayMRG

PR #380, opened 2026-05-04
support overriding ID column via mrg.id_col option

kyleam:
Various functions use "ID" as the column subject identifier, allowing it to be overridden via an argument or, in the case of one spot, hard coding it.  For convenience, allow callers to set the default via an option.

Use the "mrg." prefix because pmplots and bbr will also support specifying the ID column with this option.


R/continuous_table.R
```diff
@@ -12,7 +12,8 @@
 #'
 #' @export
 cont_table_data <- function(data, cols, by = ".total", panel = by, wide = FALSE,
-                            all_name = "all", digits = new_digits(), id_col = "ID",
+                            all_name = "all", digits = new_digits(),
+                            id_col = getOption("mrg.id_col", "ID"),
```
kyleam, line 16: @kylebaron let me know if you'd prefer I introduce a helper function rather than inlining these `getOption` calls

R/summarize-cat-chunk.R
```diff
@@ -35,7 +35,7 @@ summarize_cat_chunk <- function(data, cols, N, denom = "group") {
 #' @noRd
 summarize_cat_col <- function(name, data, D, Nchunk) {
   data <- ungroup(data)
-  pick <- select(data, "ID", level = all_of(unname(name)))
+  pick <- select(data, getOption("mrg.id_col", "ID"), level = all_of(unname(name)))
```
kyleam, line 38: Let me double check this one.

kyleam: Confirmed using alternative ID works.

```R
options("mrg.id_col" = "ALTID")
cat_data(rename(pmt_first, ALTID = ID), cols = c(SEX = "SEXf", RF = "RFf"), by = "STUDYf")
#> # A tibble: 6 × 6
#>   name  level    `12-DEMO-001` `12-DEMO-002` `11-DEMO-005` `13-DEMO-001`
#>   <chr> <chr>    <chr>         <chr>         <chr>         <chr>        
#> 1 SEX   male     10 (33.3)     18 (36.0)     29 (72.5)     23 (57.5)    
#> 2 SEX   female   20 (66.7)     32 (64.0)     11 (27.5)     17 (42.5)    
#> 3 RF    normal   30 (100.0)    50 (100.0)    10 (25.0)     40 (100.0)   
#> 4 RF    mild     0 (0.0)       0 (0.0)       10 (25.0)     0 (0.0)      
#> 5 RF    moderate 0 (0.0)       0 (0.0)       10 (25.0)     0 (0.0)      
#> 6 RF    severe   0 (0.0)       0 (0.0)       10 (25.0)     0 (0.0)   
```

Added a test to `tests/testthat/test-discrete-data.R`.

R/continuous_table.R
kylebaron: @kyleam I was going to ask the converse question: should I adopt this `getOption()` approach in `pmplots`? I think "ID" is sort of hardcoded anyway; the `getOption()` approach would make the option more transparent too.

kyleam: > the `getOption()` approach would make the option more transparent too.

Yeah, that's the main reason I prefer it (though I think either way is fine).

AFAICS the main thing using the wrapper would guard against is "oops, this one spot had a typo in the option name or fallback value".   But in my view the more obvious docs that come from removing the level of indirection are worth it.  (Also, the wrapper has a downside having an unexported function show up in the docs, which might confuse the user, or adding a function to the NAMESPACE.)

In the case of pmplots, `pm_col_id` seems to fit in well with other functions, so I could go either way.

man/cont_table_data.Rd
```diff
@@ -12,7 +12,7 @@ cont_table_data(
   wide = FALSE,
   all_name = "all",
   digits = new_digits(),
-  id_col = "ID",
+  id_col = getOption("mrg.id_col", "ID"),
   na_fill = "--",
```
kylebaron, line 16: Not specific to this PR, but what do you think about including a bulleted list of packag-wide options in `?pmtables`? I could help put that together.

kyleam: Sounds good to me.

R/continuous_table.R
```diff
@@ -143,7 +144,7 @@ pt_cont_wide <- function(data, cols,
                          all_name = "All data",
                          fun = cont_wide_fun,
                          na_fill = "--",
-                         id_col = "ID") {
+                         id_col = getOption("mrg.id_col", "ID")) {
 
   if(!missing(id_col)) {
     deprecate_warn("0.5.3", "pt_cont_wide(id_col)")
```
kylebaron: Confirmed that we need to pass `id_col` to get the deprecation warning (that's what I was expecting; just checked). 

I'm fine leaving this as is; but just wanted to ask if this could / should get handled different since the argument is deprecated? 

```r
pmtables/R (id-opt) grep deprecate *
AAAA.R:#' @importFrom lifecycle deprecate_warn
continuous_table.R:    deprecate_warn("0.5.3", "pt_cont_wide(id_col)")
continuous_table.R:    deprecate_warn("0.5.3", "pt_cont_long(id_col)")
utils.R:  stop("this function is deprecated; use `dplyr::vars` instead")

```

kylebaron (approved):
Merged in options section in `?pmtable`. Otherwise, package appears to work as expected using the option. I'm fine leaving the deprecated arguments as is; but wanted @kyleam to provide perspective on that too.

kylebaron:
EDIT: I don't know what originally motivated this with data inventory table; I wonder about an additional argument to `miss_required()` to reference the option (just for `id_col`)?

```r
  miss_required <- function(needed,pass) {
    x <- c(
      "couldn't find the {needed} column; ",
      "set the column name to '{needed}' or ",
      "pass the name as '{pass}'"
    )
    glue::glue(paste0(x,collapse=""))
  }


  if(!exists(id_col,data)) {
    emessage(miss_required(id_col, "id_col"))
    miss <- TRUE
  }

```

```r
> library(pmtables)
> data <- pmt_obs
> options(mrg.id_col = "yak")
> pt_data_inventory(data)
[pmtables] couldn't find the yak column; set the column name to 'yak' or pass the name as 'id_col'
Error: there was a problem finding required columns for the data inventory summary; please see error messages as well as '?data_inventory_chunk' help topic
```

R/continuous_table.R
kyleam: > just wanted to ask if this could / should get handled different since the argument is deprecated?

Is the main thing you have in mind just not updating it?  That'd be fine with me.  My rationale for substituting was just so that `getOption` is consistently used, but I don't think it matters either way for a deprecated arg.

kyleam:
> Merged in options section in `?pmtable`

Thanks.

> Otherwise, package appears to work as expected using the option. I'm fine leaving the deprecated arguments as is; but wanted @kyleam to provide perspective on that too.

If choice is "update to use getOption or leave be", my vote would be updating for consistency but I'm also fine leaving be, if you prefer that.

kylebaron:
@kyleam on the deprecated arguments; i'm wondering if we should just take them away? or `deprecate_stop()`?

kyleam:
> @kyleam on the deprecated arguments; i'm wondering if we should just take them away? or `deprecate_stop()`?

I see.  Looks like the deprecation warning was introduced in 2023 (0.6.0), so going to an error makes sense to me.   If you think they are unlikely to be used, I'd be in favor of just dropping them entirely and avoiding the `deprecate_stop` step.

kylebaron:
>I'd be in favor of just dropping them entirely and avoiding the deprecate_stop step.

I'd prefer to do this too; let's drop.

kyleam:
> I'd prefer to do this too; let's drop.

Dropped in d4ea7fe.  Heads up that I also removed lifecycle from `Imports` because there were no other uses, but let me know if you'd like me to add it back.

kyleam:
> I wonder about an additional argument to `miss_required()` to reference the option (just for `id_col`)?

Done in cdad68b.  Please feel free to adjust (or drop).

```R
data <- pmt_obs
options(mrg.id_col = "yak")
pt_data_inventory(data)
#> [pmtables] couldn't find the yak column; set the column name to 'yak' or pass the name as 'id_col'or set the mrg.id_col option
#> Error: there was a problem finding required columns for the data inventory summary; please see error messages as well as '?data_inventory_chunk' help topic
```

kylebaron (approved):
Thanks for making the additional changes.

PR #381, opened 2026-05-04
Document global options

kylebaron:
Add list of global options used by `pmtables`; accessible via `?pmtables`.

One option was inadvertently listed as `pmtables.text.width`; I think this was supposed to be `textwidth`; makes more sense. Update highly unlikely to affect anything.

### DIFF
--- a/R/preview-standalone.R
+++ b/R/preview-standalone.R
@@ -258,7 +258,7 @@ st_aspng <- function(x,
                      stem = "pmt-standalone-preview",
                      dir = tempdir(),
                      font = "helvetica",
-                     textwidth = getOption("pmtables.text.width", 6.5),
+                     textwidth = getOption("pmtables.textwidth", 6.5),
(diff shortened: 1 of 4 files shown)


R/preview-standalone.R
```diff
@@ -258,7 +258,7 @@ st_aspng <- function(x,
                      stem = "pmt-standalone-preview",
                      dir = tempdir(),
                      font = "helvetica",
-                     textwidth = getOption("pmtables.text.width", 6.5),
+                     textwidth = getOption("pmtables.textwidth", 6.5),
                      border = getOption("pmtables.image.border", "0.2cm 0.7cm"),
```
kylebaron, line 262: I'm 99.9% sure this should be `textwidth`; going to update.

kyleam (approved):
New text looks good to me.  Grepping for `getOption`, didn't spot any missing ones.

Issue #388, opened 2026-07-28
thick lines in stable_long outputs


callistosp:
When creating a longtable using stable_long() in combination with st_span() and st_panel(), the internal parsing seems to misplace some header commands. This results in two issues in the generated LaTeX:
 - An empty \endhead is placed immediately after \begin{longtable}.
 - An extra \hline is inserted immediately after the true \endhead, causing it to stack with the header's terminating \hline. This renders as an unexpectedly thick border line at the top of the table body in the compiled PDF.

Minimal Reproducible Example:
```
library(dplyr)
library(pmtables)

# 1. Prepare summary data
df_summ <- mtcars %>%
  mutate(cyl_group = paste(cyl, "Cylinders")) %>%
  group_by(cyl_group, gear) %>%
  summarize(
    mpg = sprintf("%.1f", mean(mpg)),
    disp = sprintf("%.1f", mean(disp)),
    hp = sprintf("%.1f", mean(hp)),
    .groups = "drop"
  )

# 2. Build the pmtables pipeline
tab_tex <- df_summ %>%
  st_new() %>%
  st_panel("cyl_group") %>%
  st_span("Performance Metrics", mpg:hp) %>%
  stable_long(
    lt_cap_text = "Motor Trend Car Road Tests",
    lt_cap_label = "tab:mtcars"
  )
  

# 3. Print to console to inspect the LaTeX output
cat(as.character(tab_tex), sep = "\n")

# 4. verify thick lines visually
st2report(tab_tex)
```

Observed Output Snippet:
```
\renewcommand{\arraystretch}{1.3}
\setlength{\tabcolsep}{5pt} 
\setlength{\extrarowheight}{0em}
\begin{longtable}{llll}
\endhead    % <--- Issue 1: Misplaced \endhead at the very beginning
\hline
\multicolumn{4}{r}{\footnotesize{continued on next page}}
\endfoot
\hline
\endlastfoot
\caption{Motor Trend Car Road Tests \label{tab:mtcars}} \\
\hline
\multicolumn{1}{c}{} & \multicolumn{3}{c}{Performance Metrics} \\
\cmidrule(lr){2-4}
gear & mpg & disp & hp \\
\hline
\endfirsthead
\hline
\multicolumn{1}{c}{} & \multicolumn{3}{c}{Performance Metrics} \\
\cmidrule(lr){2-4}
gear & mpg & disp & hp \\
\hline
\endhead
[...]

PR #389, opened 2026-08-07
Remove extra hlines in longtable

kylebaron:
There were some extraneous hlines in longtable output. Report in #388


callistosp (changes requested):
after installing this new version and running the same example in Issue #388 it does not appear that the fix is sufficient to address the issue. There is now one bolded line between 4 and 6 cylinders, and the final line only covers half the table length.

Image

kylebaron:
Can you build with `ntex = 3`? you have to build at least 3 times to get all the lines to render.

callistosp (approved):
You are right @kylebaron, great suggestion. after adding `ntex=3` the table looks as expected. Thanks for getting this addressed so quickly!

Image

kyleam (approved):

PR #391, opened 2026-09-01
Fix longtable hline to work with single- or multi-page outputs

kylebaron:
A change was made in #389 to remove extra hline at the end of longtable. That change worked fine for pdf outputs, but we don't see the bottom table line when rendering to standalone preview. 

This PR makes another update that I believe works with both pdf and png outputs. 

1. Remove `\\hline` from `\\endlasthead`
2. Force `\\hline` before closing the longtable
3. There is also a `\\` added for the continuation statement

```r
library(dplyr)
library(pmtables)


tab <- stdata() %>% st_new() %>% st_notes("dfd") %>% st_notes_detach()
tab <- stable_long(tab)

st_as_image(tab, ntex = 3)
st2report(tab, ntex = 3)

a <- stable_save_image(
  tab,
  dir = tempdir(),
  stem = "foo",
  ntex = 3
)

file.show(a)

```

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -49,3 +49,4 @@ vignettes/rename.Rmd
 ^pkgdown$
 ^\.positai$
 ^\.claude$
+^CLAUDE\.md$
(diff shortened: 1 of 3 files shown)


kyleam:
This change triggers a test failure in the regression tests added for gh-389.  I think d5c3f3f is a reasonable update for the tests, but please take a look.

kyleam (approved):
Testing the demo script here and the one in gh-388 (adding `ntex=3`), outputs look good/expected.

PR #392, opened 2026-09-01
Use latexmk to preview

kylebaron:
This PR changes the pdf build path for table previews to use `latexmk` instead of calling `pdflatex` directly. `st_aspdf()` now invokes `latexmk -pdf`, and `st2article()` / `st2report()` build the standalone document with `latexmk -pdf -halt-on-error`. Because `latexmk` reruns the tex engine as many times as needed to resolve references and settle table borders, the document is built exactly once and the `ntex` argument is no longer used for pdf output — it is retained for backward compatibility (passing it to `st2article()`/`st2report()` now signals a warning) and is still honored by `st_aspng()`, which continues to build with `latex` so `dvipng` has a dvi file to convert.


kyleam:
CI failures should be resolved with the commit I just pushed (f89cef2).

kyleam:
I think using `latexmk` so that the user doesn't need to worry about specifying `ntex` is great.

> [ntex] is still honored by`st_aspng()`, which continues to build with `latex` so `dvipng` has a dvi file to convert

I wonder whether it's worth going farther and using it here too.

What do you think about something like what I just pushed to the [latexmk-all branch](https://github.com/metrumresearchgroup/pmtables/compare/latexmk-all)?  Based on quick testing, that seems to work (though I'd definitely want you to put it through the paces on your end).

I can open that as a PR if you think that's worth pursuing.  If not, I'll continue with the review here.

kylebaron:
@kyleam sure; I can give it a try. I don't know a lot about the toolchain; if there's a path to get rid of ntex all together, that'd be great.

kylebaron:
@kyleam the latexmk-all branch works fine for me for generating standalone png, including png in html outputs, etc. As far as I can tell, works fine. 

Could you roll out this broader solution, fully deprecating `ntex`?

kyleam:
> the latexmk-all branch works fine for me for generating standalone png, including png in html outputs, etc. As far as I can tell, works fine.

Great, thanks for testing.

> Could you roll out this broader solution, fully deprecating `ntex`?

Opened PR for that branch: gh-393

kylebaron:
Closing this to pursue #393 instead.

PR #393, opened 2026-09-10
use latexmk to automate multiple builds

kyleam:
Several functions expose an `ntex` argument that callers can override when more than one build is required to reach the final output state. Switch to building with `latexmk` underneath so that the user doesn't need to worry about passing `ntex`.

---

This PR replaces gh-392.

https://github.com/metrumresearchgroup/pmtables/pull/392#issuecomment-5608126803


kylebaron (approved):
LGTM; tested both pdf and png workflows; documentation appears up to date.

